PR #6, opened 2021-03-01
read linelist, compute line profiles

ajwheeler:
This might not be ready to merge, but I thought it would be better to have this thread to talk about it while it's in progress.  I also want to see if the automated testing works.

It's very minimal.  It just parses (a few columns of) a particular linelist format (documented here http://kurucz.harvard.edu/linelists.html).  I think these are all I need for a minimum viable line opacity calculation, but I might be going back and parsing more cols later.


mabruzzo:
To answer your original question: yes I think it's fine to place the responsibility on the user for being consistent about how they supply their line lists.

> I merged, had no conflicts to resolve but committed anyway. Then I realized I hadn't pulled from origin, did that, then merged again.

Ah, that makes sense.

> Do you have thoughts about merge vs rebase?

In principle, I have a preference for rebasing - I think it tends to lead to nicer, easier to read histories (especially in large projects with lots of contributors). In practice, for larger pull requests merging can sometimes be easier (especially when you have branches depending on other branches).

I'm ready to merge this in whenever

PR #9, opened 2021-03-18
WIP: line opacities and synthesis

ajwheeler:
This is not yet working.  This pull request is to track its progress.


ajwheeler:
Sorry about the excessive committing.

src/line_opacity.jl
```diff
@@ -0,0 +1,147 @@
...
+- `total_n_density` the number density of non-electron particles in cm^-3
+- `abundances` should be a `Dict` mapping species to absolute abundances between 0 and 1.
+- `window_size` (optional, default: 20), the maximum distance from the line center at which line 
+opacities should be calculated in included.
+"""
+function line_opacity(linelist, wls, temp, electron_density, total_n_density, abundances::Dict, 
+                      atomic_masses::Dict, partition_fns::Dict, ionization_energies::Dict
+                      ; window_size=20.0)
+    #I don't love passing all these dictionaries arround, but I also don't really want to access 
+    #them as globals.  What do you think?  Maybe globals for some of them wouldn't be so bad?
```
mabruzzo, line 17: In general, I think it's better to pass things around explicitly (otherwise, I think the code's control flow and dependencies become a lot harder to reason about - especially if you've never seen it before). 

Maybe you could use globals to define default arguments? Or maybe you could package abundances & abundances & partition functions & ionization energies into a separate data structure?

ajwheeler: What do you think about turning parition_funcs and ionization_energies into functions rather that dicts. I think it's reasonable to expect that users won't want to replace them, in which case they "want" to be global, but module-level functions don't have the same confusing state problems that module level variables do.

That's just spitballing, I'm still brainstorming while I try to get a minimal example working.

ajwheeler:
Still lots to do (ctrl F TODO in the changelog, plus fixing the line and [possibly] continuum opacities), but I'm comiting because this now works to actually produce a spectrum.

ajwheeler:
I've fixed a "small" stat mech problem with calculating the line cross-sections.  It's conceptually important obviously, but it was an order 10^-3 change for many lines (although it's much larger for some).  Anyway it's fixed now.  

I also tracked down (finally) the bug that was making the line opacity way to small.  It was a unit problem, obviously.  Check out this comparison to the sun's spectrum for a small arbitrary wavelength patch:
image
image

The fact that the lines are broader is physical, i.e. not due to the line spread function, which is roughly 0.01 Å for this data.  I'm not sure exactly what is the culprit, but I suspect it's that I haven't added "microturbulence", which is really just a fudge factor that inflates the doppler broadening equally for all lines at all depths.  It *could* be that I'm taking the pressure broadening parameters from the line list, but I kind of doubt it.

Here's a larger chunk of spectrum
image
That took about 5 mins to compute. So, extrapolating, a full optical spectrum would be about 30 mins.  Yikes.  I'm sure we can get that way down, though.

ajwheeler:
I've implemented a microturbulent velocity parameter. Here's that same piece of spectrum with xi=2km/s, which is a pretty typical value.  It's definitely still not completely right, but it remains to be seen how much of that is our problem vs 1d LTE being wrong.

image

ajwheeler:
I've fixed the last little TODOs.  This is ready for review.

src/line_opacity.jl
mabruzzo: That could potentially work. I don't really see any problems with making them functions.

I do think it's important that we make it easy to replace these things for 2 reasons:
1. Our partition functions only include data for Temperatures up to 10,000 K and they don't have cutoff energy levels. Therefore, we will need different partition function data for hot stars
2. When comparing to MOOG or Turbospectrum, we may want to directly reuse their partition functions/ionization energies for the most fair comparison.

I'm a little concerned that hardcoding in a single function could make debugging issues in either of the above 2 cases much more difficult. I think I still have a slight preference for explicitly passing in this data (in some kind of atomic/molecular data struct). At the same time, I recognize that I have a tendency to future proof and this may not be necessary. I'm happy to defer to your preference on this.

mabruzzo: I think I've got a potential compromise. What if you make the partition function/ionization energy function configurable at the global module level?

I'm finding it hard to describe exactly what I'm thinking, so I'll use an example. Imagine that we had 2 or 3 different partition function implementations. For the sake of example, let's refer to the implementations as
- ``bc16_partition_func``
- ``moog_partition_func``
- ``turbospec_partition_func``
The idea would be to define some function at the top level of the module called something like ``sssynth_partition_func`` that would wrap one of implementations. 

I don't love it, but this might address my concerns about switching between implementations and your concerns about passing around too many arguments. It's just an idea. Feel free to do whatever you think is best (preferably in a separate PR)

.gitignore
```diff
@@ -1,4 +1,5 @@
 *.ipynb_checkpoints/
 *.zhistory
 Manifest.toml
-*.jl~
\ No newline at end of file
+*.jl~
+misc/Playground.ipynb
```
mabruzzo, line 5: Out of curiosity, what is ``misc/Playground.ipynb``? Is it just some scratch space?

ajwheeler:
Exactly, it’s where I made those plots. Do you know how to “dev” this
package from a different environment, btw? It’s useful for informal
testing.

RE module level configuration, I’ll have to think about it. I’m still
worried that a mutable global will make it less optimizable.

What if we have the default partition functions  as a module level
immutable global and pass it to the top level functions as a default
parameter? Obviously we would also expose the functionality to parse
partition data in whatever formats we choose, but the user would be able to
pass any Dict of callables in principle (or whatever other informal
interface we settle on).

.gitignore
ajwheeler: I foolishly replied to this in my email, see below

src/line_opacity.jl
ajwheeler: I accidentally replied to the main thread. See below.

test/runtests.jl
```diff
@@ -54,11 +85,57 @@ end
...
+    end
+
+    @testset "number densities" begin
+        abundances = Dict("C"=>1e-8, "H"=>0.9)
+        nₜ = 1e15
+        n = SSSynth.per_species_number_density(nₜ, nₜ*1e-3, 4500.0, abundances)
+        @test Set(keys(n)) == Set(["C_I", "C_II", "C_III", "H_I", "H_II"])
+        @test n["C_III"] < n["C_II"] < n["C_I"] < n["H_II"] < n["H_I"]
+        @test n["C_III"] + n["C_II"] + n["C_I"] ≈ abundances["C"] * nₜ
+        @test n["H_II"] + n["H_I"] ≈ abundances["H"] * nₜ
```
mabruzzo, line 132: Dumb question, what kind of tolerance does the `≈` symbol use? This isn't super important

test/runtests.jl
```diff
@@ -54,11 +85,57 @@ end
...
+    end
+
+    @testset "trapezoid rule" begin
+        pdf(x) = exp(-1/2 * x^2) / sqrt(2π)
```
mabruzzo, line 136: Do you think it might be useful just to add a comment saying that you are integrating over a normal distribution? Or do you think that's unnecessary?

mabruzzo (changes requested):
This looks fantastic! As someone with basically no background in this, I can follow the logic of the code really well.

I've made a suggested a handful of changes. I'm sorry that they are so nitpicky; use your digression.

mabruzzo:
> Exactly, it’s where I made those plots. Do you know how to “dev” this
> package from a different environment, btw? It’s useful for informal
> testing.

I have no idea how to do that.

> RE module level configuration, I’ll have to think about it. I’m still
> worried that a mutable global will make it less optimizable.

I don't think I was clear. I agree that the the mutable global is a bad idea. The idea is to lock in the choice of implementation when you are first loading in the module so that the optimizations can be applied to your choice. In C you would accomplish something like this with a macro or an `ifdef` statement. In Julia you could probably accomplish this more dynamically. Again I'm not totally sold on this either...

> What if we have the default partition functions  as a module level
> immutable global and pass it to the top level functions as a default
> parameter? Obviously we would also expose the functionality to parse
> partition data in whatever formats we choose, but the user would be able to
> pass any Dict of callables in principle (or whatever other informal
> interface we settle on).

Aren't we already doing this?

mabruzzo:
On the topic of speed, I have a bunch of optimization ideas. I am a little curious to profile this and see the molecular lines implementation before I start pursuing these ideas.

I realize I'm probably stating the obvious here, but the most obvious optimization is in `line_opacity`. Currently, we're doing a lot of unnecessary work for wavelengths more than a distance of `window_size` away from a given line. For large spectra (e.g. over the full visible range), I expect that this will dominate. For small spectra my lackluster implementation of the continuum opacities may dominate.

src/line_opacity.jl
```diff
@@ -1,28 +1,80 @@
...
+
+Calculate the opacity coefficient, α, from all lines in `linelist`, at wavelengths `wls`.
+
+other arguments:
+- `temp` the temerature in K
+- `window_size` (optional, default: 20), the maximum distance from the line center at which line 
+opacities should be calculated in included.
+- `vmic` is the microturbulent velocity in cm/s
+"""
+function line_opacity(linelist, wls, temp, n_densities::Dict, atomic_masses::Dict,
```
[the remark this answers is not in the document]
ajwheeler: Correct. Will do.

ajwheeler:
I think you are completely correct.  Ideally I’d like to have a dynamically
set window size but I need to think a bit about how to do it. Probably the
(product of) number density and log gf are enough, but we might want to
consider broadening parameters as well.

src/line_opacity.jl
```diff
@@ -1,28 +1,80 @@
...
-function line_profile(temp::F, atomic_mass::F, line::NamedTuple, wls::AbstractVector{F}
+function line_profile(temp::F, atomic_mass::F, ξ, line::NamedTuple, wls::AbstractVector{F}
                      ) where F <:  AbstractFloat
     #work in cgs
     λs = wls .* 1e-8 
     λ0 = line.wl * 1e-8
 
     #doppler-broadening parameter
-    Δλ_D = λ0 * sqrt(2kboltz_cgs*temp / atomic_mass) / c_cgs
+    Δλ_D = λ0 * sqrt(2kboltz_cgs*temp / atomic_mass + ξ^2) / c_cgs
```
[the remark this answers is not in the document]
ajwheeler: It does! https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/#man-numeric-literal-coefficients

Note that it modifies the operator precedence.  This sounds like it would cause bugs, but it hasn't happened to me in practice.

src/synthesize.jl
```diff
@@ -0,0 +1,195 @@
...
+    κ += ContinuumOpacity.Hminus_ff.(nH_I_div_U, nₑ, νs, ρ, T)
+    #TODO uncomment when fixed
+    #κ += ContinuumOpacity.H2plus_bf_and_ff.(nH_I_div_U, number_densities["H_II"], νs, ρ, T)
+    
+    #He continuum opacities
+    nHe_II = number_densities["He_II"]
+    nHe_II_div_U = nHe_II / partition_funcs["He_II"](T)
+    κ += ContinuumOpacity.He_II_bf.(nHe_II_div_U, νs, ρ, T)
+    #TODO fix
+    #κ += ContinuumOpacity.He_II_ff.(nHe_II, nₑ, νs, ρ, T)
```
mabruzzo, line 157:
```suggestion
    #κ += ContinuumOpacity.He_II_ff.(number_densities["He_III"], nₑ, νs, ρ, T)
```
Out of curiosity why is this commented but `ContinuumOpacity.He_II_bf` is not commented?

src/synthesize.jl
```diff
@@ -0,0 +1,195 @@
...
+"""
+function total_continuum_opacity(νs::Vector{F}, T::F, nₑ::F, ρ::F, number_densities::Dict, 
+                                 partition_funcs::Dict) where F <: AbstractFloat
+    κ = zeros(F, length(νs))
+
+    #Hydrogen continuum opacities
+    nH_I = number_densities["H_I"]
+    nH_I_div_U = nH_I / partition_funcs["H_I"](T)
+    κ += ContinuumOpacity.H_I_bf.(nH_I_div_U, νs, ρ, T) 
+    κ += ContinuumOpacity.H_I_ff.(nH_I, nₑ, νs, ρ, T)
```
mabruzzo, line 146:
```suggestion
    κ += ContinuumOpacity.H_I_ff.(number_densities["H_II"], nₑ, νs, ρ, T)
```

mabruzzo (commented):
Sorry, these 2 changes are a consequence of my mistakes. Between this and the Saha equation bugfix, I think this should hopefully resolve the issues with the continuum spectrum.

I was updating documentation and I realized that `ContinuumOpacity.H_I_ff` and `ContinuumOpacity.He_II_ff` incorrectly list `nH_I` and `nHe_II` as their respective first arguments. In reality, the first arguments should actually be `nH_II` and `nHe_III`. I'm issuing a PR later to day that fixes this issue in the definitions of these functions and clarifies the documentation.

This issue relates to the counter-intuitive naming convention for free-free interactions that we talked about previously. In this case, 
- The H I free-free interaction corresponds to `photon + electron + H II ->  electron + H II`
- The He II free-free interaction corresponds to `photon + electron + He III ->  electron + He III`

src/line_opacity.jl
```diff
@@ -1,28 +1,80 @@
...
+
+    (π*e^2/mₑ/c) * (λ^2/c) * 10^log_gf 
+end
+
+"""
+    line_profile(temp, atomic_mass, ξ, line, wl)
 
 The line profile, ϕ, at wavelengths `wls` in Ångstroms.
-`temp` should be K, `atomic_mass` should be in g.
+`temp` should be K, `atomic_mass` should be in g, `ξ` should be in cm/s,
```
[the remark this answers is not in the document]
ajwheeler: Done.

src/line_opacity.jl
```diff
@@ -1,28 +1,80 @@
...
+
+        #cross section
+        σ = sigma_line(line.wl, line.log_gf)
+
+        #stat mech quatities
+        #number density of particles in the relevant excitation state
+        boltzmann_factor = exp(- line.E_lower / kboltz_eV / temp)
+        n = n_densities[line.species] * boltzmann_factor / partition_fns[line.species](temp)
+        #the factor (1 - exp(hν₀ / kT))
+        levels_factor = 1 - exp(-c_cgs * hplanck_cgs / (line.wl * 1e-8) / kboltz_cgs / temp)
```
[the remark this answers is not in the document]
ajwheeler: Yup!

src/line_opacity.jl
```diff
@@ -1,28 +1,80 @@
 """
-    line_profile(temp, atomic_mass, line, wl)
+    line_opacity(linelist, wls, temp, n_density, abundances; window_size)
+
+Calculate the opacity coefficient, α, from all lines in `linelist`, at wavelengths `wls`.
```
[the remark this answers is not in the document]
ajwheeler: Sorry this is sloppy naming on my part.  This function does indeed return alpha.  I will address.

>To be totally honest, we should probably decide on a set of naming conventions for these quantities and settle on a set of variables and add these to the README. (Maybe I'll open an issue where we can discuss this)
Totally agree.  I think we should do this as part of the refactor.,

test/runtests.jl
ajwheeler: I believe it's machine precision by default but I'm not sure.  It's definitely below the precision we are going for.

test/runtests.jl
ajwheeler: Will do.

src/synthesize.jl
ajwheeler: Done.

src/synthesize.jl
ajwheeler: Done.

src/synthesize.jl
```diff
@@ -0,0 +1,195 @@
+using SpecialFunctions: expint
+import ..ContinuumOpacity
+
+"""
+    synthesize(atm, linelist, λs, [metallicity], [alpha]; abundances=Dict())
```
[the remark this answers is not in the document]
ajwheeler: Good point. Done.

src/line_opacity.jl
```diff
@@ -1,28 +1,80 @@
 """
-    line_profile(temp, atomic_mass, line, wl)
+    line_opacity(linelist, wls, temp, n_density, abundances; window_size)
```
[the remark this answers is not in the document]
ajwheeler: Done.

ajwheeler:
After all these changes (most notably the fixed saha equation), the lines are more-or-less the same as before.
image

The two continuum opacity sources which are causing problems at this point are H2 (as you know), and He II ff which goes outside of the interpolator bounds with log(u) values of 2-2.5.

The last thing is that we still get a big discontinuity in the continuum from the H I 
image
Without it, we see this
image


I feel like I'm forgetting something because I accidentally deleted this comment, but I can't remember what...

mabruzzo:
I'll try to take one last look at the PR later this evening and then I'll merge it. Are the tests failures just indicating that you aren't improving test coverage?

That's good to know about the continuum opacities. 
- I've got a partially working implementation of H2+ opacities, but that won't change much about this plot (it's pretty subdominant).
- I took a quick look at the implementation, and yeah you're definitely right the He lI free free absorption is breaking because log₁₀(γ²) doesn't have tabulated data above a value of 2. This corresponds to a lower temperature limit of Z^2*1578.6 K. I have an alternative data source in mind that can remedy this situation. But, I might punt on this for now since He II free-free absorption is probably subdominant to a bunch of missing opacities.

When you have a chance, could you maybe share the code snippet that you using to generate the continuum spectrum? I really just need the 2 lines of code necessary to read the atmospheric data and then execute `synthesize` (you don't have to worry about the plotting code). Tonight or tomorrow night I might try stepping through layers of the atmosphere and watching how the opacity changes. Maybe that will reveal some weird dependence in the H I continuum opacity implementation that is not otherwise obvious in the existing test.

ajwheeler:
The test failures are the fact that test coverage has gotten worse, yes.  I'm putting that off til after the big refactor.

To generate and plot a continuum, you can do:
```julia
atm = SSSynth.read_model_atmosphere("../test/data/sun.krz");
wls = 3000:10.0:9000
cntm = SSSynth.synthesize(atm, [], wls)
plot(wls, cntm.flux)
```
If you tweak this code, note that it will break if wls contains ints rather than floats.

This is the absorption coefficient alpha at each layer in the atmosphere (the top layer is in red). I already had this plot from when I was debugging the line absorption.  You can access alpha with `cntm.alpha` if you run the above code. The discontinuity does seem to be coming from the hot inner atmosphere.
image

mabruzzo:
Cool, thanks! I'll look at it more tomorrow night.

I just want to make sure that we're on the same page that the discontinuity in the opacity is a real effect - it occurs right at the wavelength where photons are no longer energetic enough to ionize electrons. Some notable examples are:
- 912 Angstroms (for n =1). This is the Lyman limit
- 3746 Angstroms (for n =2). This is the Balmer limit
- 8029 Angstroms (for n =3). This is the Paschen limit
- 14588 Angstroms (for n = 4). This is the Brackett limit

I suspect that there might be an issue with the relative normalization of the continuum opacities. Relatedly, do you if the normalization of the continuum is the right order of magnitude? And if it is closer to the expected value in any given location? Or is that difficult data to access?

mabruzzo:
My last thought is, are we sure this isn't a problem with the y-axis limits? It seems like it could have the same shape as the [continuum spectrum for the sun provided on wikipedia](https://upload.wikimedia.org/wikipedia/commons/0/0d/EffectiveTemperature_300dpi_e.png).

PR #11, opened 2021-03-27
Bugfix: A factor of 2 was missing from the Saha equation

mabruzzo:
A factor of 2 was missing from the Saha equation. Basically we were underestimating the number-densities of ionized species.

In addition to fixing this bug, I introduced the test to test/runtests.jl that helped me realize that there was an issue. This test checks the saha function against the analytic solution for a pure hydrogen atmosphere.

I think that this is probably going to resolve my outstanding issues with hydrogenic free-free opacity. But I wanted to submit this PR separately since it's going to affect the line opacities (I'm curious if it will improve agreement in PR #9 )

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -6,17 +6,56 @@ using Test
     @test SSSynth.ionization_energies["H"] == [13.5984, -1.000, -1.000]
     @test SSSynth.ionization_energies["Ru"] == [7.3605, 16.760, 28.470]
     @test SSSynth.ionization_energies["U"] == [6.1940, 11.590, 19.800]
-end 
+end
 
-@testset "saha" begin s = [SSSynth.saha(SSSynth.ionization_energies["N"], 
-                      [SSSynth.partition_funcs["N_I"], 
-                       SSSynth.partition_funcs["N_II"], 
-                       SSSynth.partition_funcs["N_III"]], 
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -6,17 +6,56 @@ using Test
...
+    @testset "pure Hydrogen atmosphere" begin
+        nH_tot = 1e15
+        # specify χs and Us to decouple this testset from other parts of the code
+        χs, Us = [SSSynth.RydbergH_eV, -1.0], [T -> 2.0, T -> 1.0]
+        # iterate from less than 1% ionized to more than 99% ionized
+        for T in [3e3, 4e3, 5e3, 6e3, 7e3, 8e3, 9e3, 1e4, 1.1e4, 1.2e4, 1.3e4, 1.4e4, 1.5e5]
+            nₑ = electron_ndens_Hplasma(nH_tot, T, 2.0)
+            weights = SSSynth.saha(χs, Us, T, nₑ)
+            @test length(weights) == 2
+            @test (weights[1] + weights[2]) ≈ 1.0 rtol = 1e-15
```
ajwheeler, line 44: I wasn't familiar with this trick.  Cool!

ajwheeler (approved):

ajwheeler:
My bad!

PR #13, opened 2021-03-28
Added tests for H I bf & ff opacities and improved documentation

mabruzzo:
The main purpose of this PR is to introduce tests confirming that the H I bf & H I ff opacities are computed correctly. These tests helped lead me to the bug introduced in PR #11.

For both the free-free and bound-free opacities, I wrote independent implementations based on the approaches described in Gray (2005). Because Gray (2005) both makes slightly different approximations and casts the constants into a completely separate form, I feel that these tests are a relatively robust accuracy check. I added an additional minor testset for the bound-free opacities, checking the validity of our integral approximation. 

I additionally created a test comparing the combined H I bf and H I ff opacity values to the tabulated curves from the panels of figure 8.5 from Gray (2005). The updated plots are provided below. As you can see, PR #11 largely reconciled the differences in the curves that I had previously shared. While some discrepancy still remains, I'm not concerned about it all; the individual tests (described in the last paragraph) indicate that there is reasonable agreement. The remaining discrepancy is plausibly described by differences in the approximations, which are:
- Gray (2005) suggests that the summation in the bound-free opacity can be replaced by an integral at smaller n values than Kurucz (1970); note that above ~4.45 microns Kurucz (1970)'s approach uses in a comparable or potentially more aggressive manner. This helps explain why our approach always gives smaller values.
- Gray (2005) suggests the use of a much older/analytic approximations for both sets of gaunt factors. I think this generally means that their results are less accurate.

If you want a more detailed description about the changes that were made, I would encourage you to take a look at the commit messages (they're fairly detailed).

**Future Work on Continuum Opacities:**
I've still got to fix the H₂⁺ ff and bf opacities, but I'll do that in a separate branch. After I do that, I'll start refactoring the continuum opacities so that we achieve better performance. At this point I think that the dominant continuum opacities are robust enough that we should be able to start generating realistic spectra.

**Plotted Opacities:**
Panel B (T = 6428 K, Pₑ = 10^1.77 dyne cm⁻²):
test_b

Panel C (T = 7715 K, Pₑ = 10^2.50 dyne cm⁻²):
test_c

Panel D (T = 11572 K, Pₑ = 10^2.76 dyne cm⁻²):
test_d

### DIFF
--- a/src/continuum_opacity/hydrogenic_bf_ff.jl
+++ b/src/continuum_opacity/hydrogenic_bf_ff.jl
@@ -148,8 +148,6 @@ integral.
 
 # Arguments
 - `Z::Integer`: Z is the atomic number of the species (e.g. 1 for H I or 2 for He II)
-- `nmax_explicit_sum::Integer`: The highest energy level whose opacity contribution is included in
-   the explicit sum. The contributions from higher levels are included in the integral.
 - `nsdens_div_partition::Flt` is the total number density of the species divided by the species's
    partition function.
 - `ν::Flt`: frequency in Hz
@@ -157,14 +155,16 @@ integral.
 - `T::Flt`: temperature in K
 - `ion_energy::AbstractFloat`: the ionization energy from the ground state (in eV). This can be 
    estimated as Z²*Rydberg_H (Rydberg_H is the ionization energy of Hydrogen)
+- `nmax_explicit_sum::Integer`: The highest energy level whose opacity contribution is included in
+   the explicit sum. The contributions from higher levels are included in the integral.
 - `integrate_high_n::bool`: When this is `false`, bf opacity from higher energy states are not
    estimated at all. Default is `true`.
 
 # Notes
 This follows the approach described in section 5.1 of Kurucz (1970).
 """
-function hydrogenic_bf_opacity(Z::Integer, nmax_explicit_sum::Integer, nsdens_div_partition::Flt,
-                               ν::Flt, ρ::Flt, T::Flt, ion_energy::Flt,
+function hydrogenic_bf_opacity(Z::Integer, nsdens_div_partition::Flt, ν::Flt, ρ::Flt, T::Flt,
+                               ion_energy::Flt, nmax_explicit_sum::Integer,
                                integrate_high_n::Bool = true) where {Flt<:AbstractFloat}
     ionization_freq = _eVtoHz(ion_energy)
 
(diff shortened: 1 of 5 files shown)


ajwheeler:
This is amazingly complete!

src/continuum_opacity/hydrogenic_bf_ff.jl
```diff
@@ -148,23 +148,23 @@ integral.
...
 - `integrate_high_n::bool`: When this is `false`, bf opacity from higher energy states are not
    estimated at all. Default is `true`.
 
 # Notes
 This follows the approach described in section 5.1 of Kurucz (1970).
 """
-function hydrogenic_bf_opacity(Z::Integer, nmax_explicit_sum::Integer, nsdens_div_partition::Flt,
-                               ν::Flt, ρ::Flt, T::Flt, ion_energy::Flt,
+function hydrogenic_bf_opacity(Z::Integer, nsdens_div_partition::Flt, ν::Flt, ρ::Flt, T::Flt,
+                               ion_energy::Flt, nmax_explicit_sum::Integer,
```
ajwheeler, line 167: What's the reason for changing the argument order here?  It's not a problem, just curious.

src/continuum_opacity/opacity_H.jl
```diff
@@ -183,10 +186,10 @@ function Hminus_ff(nH_I_div_partition::Flt, ne::Flt, ν::Flt, ρ::Flt,
...
     Pₑ = ne * kboltz_cgs * T
 
-    # in this case, I'm going to account for the fact that n(H I, n=1) might be slightly
-    # smaller than the entire number density of H I. This is only a difference at the highest
-    # temperatures. For the temperature range where this approximation is valid, less that
-    # 0.23% of all H I atoms are not in the ground state.
+    # Going to account for the fact that n(H I, n=1) might be slightly smaller than the entire
+    # number density of H I. This is only a difference at the highest temperatures. For the
+    # temperature range where this approximation is valid, less than 0.23% of all H I atoms are not
+    # in the ground state.
```
ajwheeler, line 192: Is there a typo in the first sentence here?

ajwheeler:
Please merge once/if you address the small comments above.

src/continuum_opacity/hydrogenic_bf_ff.jl
mabruzzo: This was changed to reflect the argument order of the revised function signature of `ContinuumOpacity.H_I_bf(nH_I_div_partition, ν, ρ, T, ion_energy = _H_I_ion_energy, nmax_explicit_sum = 8, integrate_high_n = true)`. And I revised the signature of `ContinuumOpacity.H_I_bf` so that I could easily specify `nmax_explicit_sum` and `integrate_high_n` in a test.

src/continuum_opacity/opacity_H.jl
mabruzzo: Yeah, you're totally right. Fixed.

PR #16, opened 2021-03-31
new function to apply LSF

ajwheeler:
This introduces a function that calculates the matrix which maps spectra to LSF-convolved spectra, assuming the LSF is a gaussian with width set by a constant R = lambda/Delta lambda.  It assumes that the wavelength grid is locally-linear and dense, or it will not generally be normalized correctly.  In the future, I plan to automate the process of synthesizing on a dense grid, convolving with the LSF, and downsampling automatic, since it involves a few steps.

Here's the Ca triplet (what the Gaia RVS uses), with a realistic-ish LSF imposed.
image

### DIFF
--- a/src/LSF.jl
+++ b/src/LSF.jl
@@ -1,11 +1,12 @@
 normal_pdf(Δ, σ) = exp(-0.5*Δ^2 / σ^2) / √(2π) / σ
 
 """
-Returns the matrix which transforms a spectrum to its LSF-convolved version for spectral resolution 
+Returns the matrix which a gaussian line spread function with constant spectral resolution, 
 (λ / Δλ) R.
 
 This will have weird behavior if your wavelength grid is not locally linearly-spaced.
-It is intended to be run on a fine wavelength grid, then downsampled to the observational (or otherwise desired) grid.
+It is intended to be run on a fine wavelength grid, then downsampled to the observational (or 
+otherwise desired) grid.
 """
 function constant_R_LSF(flux::AbstractVector{F}, wls, R) where F <: AbstractFloat
     #ideas - require wls to be a range object? Use erf to account for grid edges?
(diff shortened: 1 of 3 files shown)


ajwheeler:
I just realized that this is totally untenable for large spectra.  I will address by not explicitly constructing a matrix.

ajwheeler:
This now computes the convolved spectrum directly.  It's not as fast as would be idea.  (It takes about 2 mins to on the full visible with a dense wl grid, but we can look into efficient methods later.  I'm not really sure what the options are since the convolution kernel in wl-dependent.  Maybe we could transform to a basis where it's constant?  Then the normalization might be tricky...

mabruzzo:
What exactly is a LSF-convolved spectra? Does this accounting for the fact that an instrument has finite resolution?

ajwheeler:
This doesn’t account for the pixels on the detector, but it account for the line spread from the instrument optics, which generally has this functional form to first (second?) order. 

Dealing with the detector pixels is also important, but we aren’t doing this at the moment.

src/LSF.jl
```diff
@@ -0,0 +1,22 @@
+normal_pdf(Δ, σ) = exp(-0.5*Δ^2 / σ^2) / √(2π) / σ
+
+"""
+Returns the matrix which transforms a spectrum to its LSF-convolved version for spectral resolution
```
mabruzzo, line 4: could you maybe just spell out "line spread function"? Other than that, this seems great

src/LSF.jl
```diff
@@ -1,11 +1,12 @@
 normal_pdf(Δ, σ) = exp(-0.5*Δ^2 / σ^2) / √(2π) / σ
 
 """
-Returns the matrix which transforms a spectrum to its LSF-convolved version for spectral resolution 
+Returns the matrix which a gaussian line spread function with constant spectral resolution,
```
mabruzzo, line 4: I don't mean to nitpick, but there's a typo here

ajwheeler: Not only that, it was factually wrong!  thanks.

PR #18, opened 2021-04-02
Speedup

ajwheeler:
This addresses the most egregious performance problems.  I foolishly wrote code that wave quadratic in the size of the wavelength grid, fixing that halved the runtime to synthesize the Ca triplet, which is only a few tens of Angstroms, so it will be a massive difference for large wl ranges.  I've updated the profiling results at http://user.astro.columbia.edu/~aw/statprof/ to reflect that.

I'm looking for any other obvious issues now, will let you know when this is ready to be reviewed and merged.


ajwheeler:
This isn't passing because it replaces a few lines of untested code with slightly more lines of untested code.  I've verified that it generates the same synthetic spectrum, so I think it's fine.  At some point, I'll have to figure out how to write good tests for some of the line absorption functions.

ajwheeler:
I didn't find anything else obvious.  Ready for review.

src/line_opacity.jl
```diff
@@ -76,7 +85,9 @@ function line_profile(temp::F, atomic_mass::F, ξ::F, line::NamedTuple, wls::Abs
 
     #Voigt function parameters
     v = @. abs(λs-λ0) / Δλ_D
-    a = @. λs^2/(4π * c_cgs) * γ / Δλ_D
+    #technically, a depends on λ, but at the sub-percent level, which is below the precision of our
+    #voigt approximation
+    a = λ0^2/(4π * c_cgs) * γ / Δλ_D
```
mabruzzo, line 90: Do you think it would potentially make sense to document somewhere that this change right here places a constraint on `line.wl` and `window_size`? This just seems like an unexpected place that the code could break if the rest of the codebase ever gets extended for UV spectra (Maybe you could add an `assert` statement at the start of `line_absorption` with a comment explaining why it's there).

I realize that this is future proofing. If you don't think it's a concern, feel free to ignore this

mabruzzo (approved):
This seems good to me. After you decide whether my 1 comment needs to addressed feel free to merge

src/line_opacity.jl
ajwheeler: This is a good idea.  I can add a comment to the source, but should we have a central place where we record the limitations of various approximations?  I definitely don't want to have an in-code check here.

Alternatively, this is a very small performance gain, maybe I should just revert it?  I'm kind of leaning towards that.

ajwheeler: I went ahead and rolled back that approximation.

PR #19, opened 2021-04-02
don't mutate linelist when filtering it, store a new filtered version

ajwheeler:
This tiny PR fixes a small bug.


mabruzzo (approved):
I agree that this is better than mutating the existing linelist. 

Since linelists could potentially be massive (in extreme cases having upwards of 100,000 entries) and we are already assuming that it's sorted, would it maybe be better to take a slice instead (to avoid the copy)? This may be over-engineering so feel free to merge.

ajwheeler:
It’s good to keep in mind that making a copy is potentially a high-memory operation, but I want this to work in cases where the linelist has “gaps”.

PR #20, opened 2021-04-04
simplify units: use cm throughout in place of Angstroms

ajwheeler:
I also switched from `wls` to `λs` in a couple places for more consistency.


src/synthesize.jl
```diff
@@ -5,27 +5,29 @@ import ..ContinuumOpacity
...
-   before it is discarded.
-
+- `line_window` (default: 10): the farthest any line can be from the provide wavelenth range range
+   before it is discarded (in Å).
 Uses solar abundances scaled by `metallicity` and for those not provided.
 """
 function synthesize(atm, linelist, λs::AbstractVector{F}, metallicity::F=0.0; vmic=1.0, 
                     abundances=Dict(), line_window::F=10.0) where F <: AbstractFloat
+    #work in cm
+    λs *= 1e-8
```
mabruzzo, line 21: Would you mind explaining your rational as to why the arguments are specified in Å instead of cm? (I'm just curious)

Additionally I think you should be careful about modifying `λs` in-place as it produces an unexpected side-effect. I think you should either document in the docstring that this is one of the consequences of the function call or you should store the values in a new array.

mabruzzo (changes requested):

mabruzzo:
Do you know why the macOS tests failed?

src/synthesize.jl
ajwheeler: Even though I think it's easier if the code internals are all cgs, I think the user-facing/top-level function should take Angstroms since that's the standard in the field.

> Additionally I think you should be careful about modifying λs in-place as it produces an unexpected side-effect. I think you should either document in the docstring that this is one of the consequences of the function call or you should store the values in a new array.

Good catch.  I will address this.

ajwheeler:
> Do you know why the macOS tests failed?

Not sure, it looks like it timed out?  We can look into it if it happens again.

PR #21, opened 2021-04-04
Fixed/updated H2plus opacity and updated Heminus ff opacity 

mabruzzo:
I updated the implementation of `ContinuumOpacity.H2plus_bf_and_ff` so that it now works. I also updated `ContinuumOpacity.Heminus_ff` to enforce the valid wavelength range. 

Unfortunately the minimum wavelength at which we can evaluate `ContinuumOpacity.Heminus_ff` is 506.3 nm, which is right in the middle of the visible wavelength range. For now, I've commented the call to this function in `total_continuum_opacity` to avoid errors. I'm really not sure if there is an alternative implementation that we can use. I'm pretty sure that MARCS uses the same approach.

I also introduced some tests for `ContinuumOpacity.H2plus_bf_and_ff` and a new test for `ContinuumOpacity.Heminus_ff`. Here is the updated comparison against the only panel of figure 8.5 from Gray (2005) that includes non-neglible H₂⁺ ff and bf opacity.
test_b

### DIFF
--- a/src/continuum_opacity/opacity_H.jl
+++ b/src/continuum_opacity/opacity_H.jl
@@ -122,7 +122,6 @@ function Hminus_bf(nH_I_div_partition::Flt, ne::Flt, ν::Flt, ρ::Flt, T::Flt,
 end
 
 
-
 """
     Hminus_ff(nH_I_div_partition, ne, ν, ρ, T)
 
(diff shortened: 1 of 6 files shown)


src/continuum_opacity/opacity_H.jl
```diff
@@ -199,43 +198,91 @@ function Hminus_ff(nH_I_div_partition::Flt, ne::Flt, ν::Flt, ρ::Flt,
...
 
-
 """
-    H2plus_bf_and_ff(nH_I_div_partition, nHII, ν, ρ, T)
+    H2plus_bf_and_ff(nH_I_div_partition, n_HII, ν, ρ, T)
+
+Compute the combined H₂⁺ bound-free and free-free opacity κ.
 
-Compute the combined H₂⁺ bound-free and free-free opacity κ
+This uses polynomial fits from Gray (2005). that were derived from data tabulated in
```
ajwheeler, line 206: Nitpick: this period should be a comma.

src/continuum_opacity/opacity_H.jl
```diff
@@ -199,43 +198,91 @@ function Hminus_ff(nH_I_div_partition::Flt, ne::Flt, ν::Flt, ρ::Flt,
...
+
+    atomic_cross_section = begin
+        σ1 = -1040.54 + 1345.71 * logλ - 547.628 * log2λ + 71.9684 * log3λ
+        # there was a typo in Gray (2005). In the book they give the following polynomial
+        # as the fit for U₁. In reality, it is the fit for negative U₁
+        neg_U1 = 54.0532 - 32.713 * logλ + 6.6699 * log2λ - 0.4574 * log3λ
+
+        # note: Gray (2005) used the equivalent expression: σ1*10^(neg_U1 * θ) where θ = 5040/T
+        σ1 * exp(neg_U1 * β_eV)
+    end
```
ajwheeler, line 279: This isn't important, but I used to like this pattern because I think it introduced a separate scope and that local variables would be captured.  Since you pointed out that this isn't true, I'm not sure what the benefits are.  IDK it's definitely not a problem to do it this way though.

PR #29, opened 2021-04-16
Moog linelist format and linelist refactor

ajwheeler:
This adds a parser for MOOG-formatted line lists, which do not always contain broadening parameters.  The broadening parameter estimation will have to be improved in the future, but that's kind of a separate matter.  An overhaul of broadening is at the top of my todo list.

It also introduces a new type, `Line`, to replace the `NamedTuple`s  we've been using thus far.  The main advantage at this point is that it will catch more errors at compile time and it pretty prints lines in the REPL, but I think this is the way to go as we add more functionality (or even if we don't).


src/linelist.jl
```diff
@@ -55,63 +66,55 @@ function get_atoms(molecule)
...
+end
+
+"""
+Construct a `Line` without explicit broadening parameters.
+"""
+function Line(wl::F, log_gf::F, species::String, E_lower::F) where F
+    e = electron_charge_cgs
+    m = electron_mass_cgs
+    c = c_cgs
+    Line(wl, log_gf, species, E_lower, log10(8π^2 * e^2 / m / c / wl^2) + log_gf, -Inf, -Inf)
```
mabruzzo, line 89: For performance reasons, it might be advisable avoiding the use of `Inf`. I'm pretty sure that there is extra overhead in floating point operations involving `Inf`. Although, I totally understand if you decide that this optimization is not necessary.

If you wanted to do something different, you could potentially:
- introduce a flag to `Line` that indicates whether the broadening parameters are missing
- Store the broadening parameter instead of log10 of the broadening parameter
- You could replace `-Inf` with `nextfloat(log10(floatmin(F)))`. Note that the use of `nextfloat` seeks to avoid evaluation to a [subnormal](https://en.wikipedia.org/wiki/Denormal_number) value when you raise 10 to the power of this value. When `log_gamma_rad` is greater than or equal to -30.4 (-291.3), then this won't introduce any error while using 32-bint floats (64-bit floats)

mabruzzo (approved):
This looks great to me.

I just want to emphasize that you don't need to feel obliged to act on my one comment about the usage of `Inf`.

Once your resolve the merge conflict, feel free to merge this in

src/linelist.jl
```diff
@@ -55,63 +66,55 @@ function get_atoms(molecule)
     el1, el2        
 end
 
+"""
+This type represents an individual line.
+"""
+struct Line{F}
```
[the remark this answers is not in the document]
ajwheeler: Exactly

src/linelist.jl
ajwheeler: I am probably going to switch from log gamma to gamma in the near future. But also, those infs need to be replaced with a scientifically reasonable approximation no matter what, so they will be gone by the time we are ready to submit.

PR #30, opened 2021-04-16
Interpolate cntm

ajwheeler:
This interpolates the continuum opacity over a sparser wavelength grid, then interpolates it to the fine grid with linear interpolation.  This is branched off #29, so that should be merged first.

The technically messes up the balmer jump and similar bf discontinuities, but the default continuum wl grid 1 A, so I think it's within the realm of what's acceptable.  Maybe in the future we could accommodate precise bf jumps though.


src/synthesize.jl
```diff
@@ -24,12 +25,14 @@ optional arguments:
 Uses solar abundances scaled by `metallicity` and for those not provided.
 """
 function synthesize(atm, linelist, λs::AbstractVector{F}, metallicity::F=0.0; vmic=1.0, 
-                    abundances=Dict(), line_window::F=10.0, 
+                    abundances=Dict(), line_window::F=10.0, cntm_step=1.0::F,
```
mabruzzo, line 28: Can you add a line to docstring to describe `cntm_step`?

mabruzzo (changes requested):

mabruzzo (approved):

mabruzzo:
I just fixed the merge conflict. After the tests passed it can be merged in

PR #34, opened 2021-05-01
change SSSynth to Korg everywhere

ajwheeler:


ajwheeler:
Upon merging, we will have to change the repo name for the readme links and badges to not be broken.

mabruzzo (approved):

PR #40, opened 2021-05-12
Autodiff

ajwheeler:
This makes types generic enough to support autodiff.  Unfortunately, it is also necessary to define a method of `floatmax` which is defined on the relevant `ForwardDiff.Dual` subtype, e.g.
```julia
Base.floatmax(::ForwardDiff.Dual{ForwardDiff.Tag{typeof(syn), Float64}, Float64, 2}) = floatmax(1.0)
```
where `syn` is the function being differentiated, e.g. the spectrum as a function of Teff and vmic:
```julia
syn(params) = synthesize(atm, linelist, wls; metallicity=params[1], vmic=params[2]).flux
```

I will make sure this is documented at some point, and will take care of it myself in the Korg parameter estimation routines.

BTW, calculating the Jacobian in this example takes the same amount of time as synthesizing the spectrum.  Not bad!

P.S. this is branched off #32.  I will fix the diff before review.

### DIFF
--- a/src/continuum_opacity/hydrogenic_bf_ff.jl
+++ b/src/continuum_opacity/hydrogenic_bf_ff.jl
@@ -10,11 +10,11 @@ Calculates the LTE number density (in cm^-3) of a hydrogenic species at a given 
 
 # Arguments
 - `n::Integer`: The quantum number of the state
-- `nsdens_div_partition::Flt`: The total (including all states) number density of the current 
+- `nsdens_div_partition`: The total (including all states) number density of the current 
 ionization species (in cm^-3) divided by the species's partition function.
-- `ion_energy::Flt`: The minimum energy (in eV) required to ionize the species (from the ground 
+- `ion_energy`: The minimum energy (in eV) required to ionize the species (from the ground 
 state). This can be estimated as Z²*ion_energy of hydrogen or Z²*Rydberg_H.
-- `T::Flt`: Temperature
+- `T`: Temperature
 
 # Note
 This assumes that all hydrogenic species have a statistical weight of gₙ = 2*n².
@@ -22,8 +22,7 @@ This assumes that all hydrogenic species have a statistical weight of gₙ = 2*n
 This was taken from equation (5.4) of Kurucz (1970) (although this comes directly from the 
 boltzmann equation).
 """
-function ndens_state_hydrogenic(n::Integer, nsdens_div_partition::Flt, T::Flt,
-                                ion_energy::Flt) where {Flt<:AbstractFloat}
+function ndens_state_hydrogenic(n::Integer, nsdens_div_partition::Real, T::Real, ion_energy::Real)
     n2 = n*n
     g_n = 2.0*n2
     energy_level = ion_energy - ion_energy/n2
@@ -46,7 +45,7 @@ const _bf_σ_coef = SA[(0.9916,  2.719e13, -2.268e30),
  
 # this is helper function is the most likely part of the calculation to change.
 # This uses double precision to be safe about the polynomial coefficients.
-function _hydrogenic_bf_cross_section(Z::Integer, n::Integer, ν::Float64, ion_freq::Float64)
+function _hydrogenic_bf_cross_section(Z::Integer, n::Integer, ν::Real, ion_freq::Real)
     # this implements equation 5.5 from Kurucz (1970)
     # - Z is the atomic number
     # - n is the energy level (remember, they start from n=1)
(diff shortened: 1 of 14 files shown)


ajwheeler:
I forgot to mention, the new method for `floatmax` is necessary because of `SpecialFunctions.expint`.  If we implement our own exponential integral (which might be worth it for speed; this version is accurate to floating point precision), we wouldn't have to do that.

ajwheeler:
image
I've written an approximate exponential integral function.  This is 1) faster and 2) doesn't cause problems with dual numbers.

As you can see in the above plots, it's accurate to within 1%.  (This is also true for larger values of x.)

At some point, someone who is better at math than me could probably get something working which is more compact/accurate, but I think this is definitely good enough for our purposes.

src/fit.jl
```diff
@@ -0,0 +1,24 @@
...
+    mask = rand(length(wls)) .< wl_deflation_factor
+    
+    result = optimize(p->chi2(p, mask), [0.0, 0.0], NewtonTrustRegion(),
+        Optim.Options(show_trace=true, extended_trace=true, x_tol=1e-1);
+        autodiff=:forward)
+    
+    optimize(p->chi2(p, ones(Bool, length(wls))), result.minimizer, NewtonTrustRegion(),
+        Optim.Options(show_trace=true, extended_trace=true, x_tol=1e-3);
+        autodiff=:forward)
+end
```
ajwheeler, line 24: All of this is junk that I will replace.  It's completely broken I just didn't want to lose the code.

ajwheeler:
I updated this branch with rebase/force-push. I think everything worked but if there's git problems that's probably why.

ajwheeler:
This is a really ugly combo of different things, but it would be great to get it merged.  Everything except the "early fitting code" is ready for review.  What do you think is the best way to go about it?

src/continuum_opacity/hydrogenic_bf_ff.jl
```diff
@@ -164,9 +161,9 @@ integral.
 # Notes
 This follows the approach described in section 5.1 of Kurucz (1970).
 """
-function hydrogenic_bf_opacity(Z::Integer, nsdens_div_partition::Flt, ν::Flt, ρ::Flt, T::Flt,
-                               ion_energy::Flt, nmax_explicit_sum::Integer,
-                               integrate_high_n::Bool = true) where {Flt<:AbstractFloat}
+function hydrogenic_bf_opacity(Z::Integer, nsdens_div_partition::Real, ν::Real, ρ::Real, T::Real, 
+                               ion_energy::Real, nmax_explicit_sum::Integer, 
+                               integrate_high_n::Bool=true)
```
mabruzzo, line 166:
```suggestion
function hydrogenic_bf_opacity(Z::Integer, nsdens_div_partition::F, ν::F, ρ::F, T::F, 
                               ion_energy::F, nmax_explicit_sum::Integer, 
                               integrate_high_n::Bool=true) where {F<:Real}
```

src/continuum_opacity/hydrogenic_bf_ff.jl
```diff
@@ -10,20 +10,19 @@ Calculates the LTE number density (in cm^-3) of a hydrogenic species at a given
...
 
 # Note
 This assumes that all hydrogenic species have a statistical weight of gₙ = 2*n².
 
 This was taken from equation (5.4) of Kurucz (1970) (although this comes directly from the 
 boltzmann equation).
 """
-function ndens_state_hydrogenic(n::Integer, nsdens_div_partition::Flt, T::Flt,
-                                ion_energy::Flt) where {Flt<:AbstractFloat}
+function ndens_state_hydrogenic(n::Integer, nsdens_div_partition::Real, T::Real, ion_energy::Real)
```
mabruzzo, line 25:
```suggestion
function ndens_state_hydrogenic(n::Integer, nsdens_div_partition::F, T::F,
                                ion_energy::F) where {F<:Real}
```

mabruzzo (approved):
This seems good to me.

I had a few minor comments.

I also suggested a bunch of changes to the continuum opacity function signatures that make check to make sure that the individual arguments receive arguments of the same type. This is a lot closer to what I originally had and is consistent with the style of function signatures that you use in `line_opacity.jl` and 'line_list.jl`. I also think that this style makes the return type a lot more predictable and helps avoid bugs. But if there's a reason that we can't do this, just let me know.

ajwheeler:
Not that I would expect you to re-review and merge this today, but please don't merge.  I think I've identified a bug and am working to fix it.

ajwheeler:
No bug per se, but I think the "excess broadening" we've been seeing is actually continuum opacities that are too small, presumably because I've had to comment out various sources that don't have good data in the IR.

src/continuum_opacity/opacity_H.jl
```diff
@@ -207,12 +206,12 @@ This uses polynomial fits from Gray (2005) that were derived from data tabulated
...
+- `nH_I_div_partition`: the total number density of H I divided by its partition 
    function.
-- `nH_II::Float64`: the number density of H II (not of H₂⁺).
-- `ν::Float64`: frequency in Hz
-- `ρ::Float64`: mass density in g/cm³
-- `T::Float64`: temperature in K
+- `nH_II`: the number density of H II (not of H₂⁺).
+- `ν`: frequency in Hz
+- `ρ`: mass density in g/cm³
+- `T`: temperature in K
```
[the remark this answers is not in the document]
mabruzzo: To be totally honest, I can't remember. I think that I might have just stopped keeping track after a while

PR #42, opened 2021-06-13
WIP: Opacity Project bf opacities

mabruzzo:
This is very much a work-in-progress. But I wanted to post what I have so far.

I'm fairly confident that I have the code working to compute the bound-free cross sections for different metals from the Opacity Project. Basically it takes a weighted average of the cross-sections from the different electron configurations of a given ionization state (assuming LTE) and makes the correction for stimulated emission.

I'm starting to get pretty close. I still need to:
 - [ ] Add additional documentation for `_OP_bf_tabulated_cross_section`
 - [ ] Clean a lot of the debugging/test code from the file I was using and make sure this can actually be imported as a part of Korg
 - [ ] Add some test code to actually compute the opacity
 - [ ] Introduce some formal tests to make sure I'm doing things correctly. I'm thinking of comparing the bf opacities computed for H I and He II to the values returned by the functions that assume a Hydrogenic atom.


ajwheeler:
How difficult is it to mass-download the necessary opacity project data files in the right format?  Is it something the user could pass in?

mabruzzo:
> How difficult is it to mass-download the necessary opacity project data files in the right format? Is it something the user could pass in?

Currently, it's a little involved. 

The Opacity Project provides the data in an alternative file-format that would be easier for users to mass-download and we're going to need to be able parse anyway if we Fe I and Fe II opacities. But, we can't compute opacities from that data format until we can:
- manually compute the statistical weight of a given electron configuration (this would take a little work)
- Determine the energy level of the ground-state (this is probably easy). I'm pretty sure that the table tells us the energy level of each electron configuration with respect to the ionization potential. I would assume that the ground state is just the most stable state (the lowest energy), but I'm slightly concerned there could be something more to it related to electron configuration). Regardless, this is easily testable using of the data formats we can already parse.

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,561 @@
...
+
+"""
+    StateID(iSLP, iLV)
+
+`StateID` represents a given state (electron configuration) of an ion species. The arguments are
+taken from the TOPBase.
+
+# Arguments
+- `iSLP::UInt16`: An integer representing the quantum atomic numbers SLπ (total spin, orbital
+  angular momentum, and parity). The value is given by `100*(2*S+1) + 10*L + π`. Note that TOPBase
```
ajwheeler, line 21: Forgive my ignorance, why not `100*S`?

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,561 @@
...
+    iLV::UInt8
+    function StateID(iSLP, iLV)
+        @assert 99 < iSLP < 992 # this upper limit is set based on the data source
+        @assert (iSLP % 10) < 2 # trailing digit must be 0 or 1
+        @assert 0 < iLV # the upper limit is set based on the data source
+        new(iSLP, iLV)
+    end
+end
+
+Base.:(<)(a::StateID, b::StateID) = (a.iSLP < b.iSLP) || ((a.iSLP == b.iSLP) && (a.iLV < b.iLV))
```
ajwheeler, line 48: I *think* you could define this as
```julia
Base.:(<)(a::StateID, b::StateID) = (a.iSLP, a.iLV) < (b.iSLP, b.iLV)
```

although I suppose yours is faster...

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,561 @@
...
+Base.:(==)(a::StateID, b::StateID) = (a.iSLP == b.iSLP) && (a.iLV == b.iLV)
```
ajwheeler, line 49: Because this is an immutable struct, `==` falls back to `===`, which does what you want without a new method.  The same is true for `hash`, I believe.

src/continuum_opacity/opacity_metal.jl
mabruzzo: I'm speculating, but maybe S can be equal to 0?

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,561 @@
...
+
+# Notes
+The combination of this struct and an ion species (or neutral element) should be enough to fully
+specify a given electron configuration. For example, consider He I:
+- `iSLP = 100, iLV = 1` has the electron configuration: 1s²
+- `iSLP = 100, iLV = 2` has the electron configuration: 1s 2s
+- `iSLP = 100, iLV = 3` has the electron configuration: 1s 3s
+- `iSLP = 111, iLV = 1` has the electron configuration: 1s 2p
+- `iSLP = 111, iLV = 2` has the electron configuration: 1s 3p
+"""
```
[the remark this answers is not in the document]
mabruzzo: The short answer is: you don't. I don't think that the Opacity Project lists this in any of their tables.

For what it's worth, my remaining intuition from chemistry class tells me that's very unstable.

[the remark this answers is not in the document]
ajwheeler: I suspected as much

src/continuum_opacity/opacity_metal.jl
mabruzzo: Interesting, then I'll definitely drop `Base.:(==)`.

I'll try to drop my definition of `hash` too (I couldn't find any documentation about the default implementation of `hash`)

mabruzzo:
@ajwheeler  This is now ready for you to review. I've definitely cut some corners at the end, so definitely call me out where you think I need to improve stuff... I'll also make a point of looking over your previous suggestions in the next few days... (I had completely forgot about it)

Also Here are my 2 comparison plots:
H_I_comp

He_II_comp

test/continuum_opacity.jl
```diff
@@ -1,5 +1,135 @@
...
+
+# Keywords
+- `rtol`: The desired relative tolerance
+- `atol`: The desired absolute tolerance
+- `err_msg`: An optional error message that can be printed when the arrays are not close
+- `error_location_fmt`: An optional keyword used to specify a function that creates a string 
+  describing the location in the arrays where values are unequal, given a `CartesianIndex` object.
+  This should not be specified in cases where actual and reference have different shapes.
+"""
+function assert_allclose(actual, reference; rtol = 1e-7, atol = 0.0, err_msg = nothing,
```
ajwheeler, line 22: Does it make sense to move this into its own file in `/test`?  Seems like it might as well be used throughout the test suite if it's defined.

test/continuum_opacity.jl
```diff
@@ -538,8 +668,56 @@ end
...
+        hydrogenic_α_OP = OP_compare.calc_hydrogenic_bf_absorption_coef(λ_vals, T, ndens_species,
+                                                                        species_name;
+                                                                        use_OP_data = true)
+        # compute the absorption coefficients using our function for hydrogenic atoms
+        hydrogenic_α_dflt = OP_compare.calc_hydrogenic_bf_absorption_coef(λ_vals, T, ndens_species,
+                                                                          species_name;
+                                                                          use_OP_data = false)
+
+        λ_comp_intervals = OP_compare._λ_comp_intervals(species_name)
+        comp_ind = _contained_in_inclusive_intervals(λ_vals, λ_comp_intervals)
```
ajwheeler, line 712: Does this work?
```suggestion
        comp_ind = map(λ_vals) do λ
            any(λ_comp_intervals[:, 1] .<= λ .<= λ_comp_intervals[:, 2])
        end
```

ajwheeler (commented):
Ok, I've taken a pass through this.  I was a bit intimidated by `opacity_metal.jl` in particular, so please tell me if there are any places where you would especially like attention.  The parsing code was had to, well, parse without knowledge of the data so I especially neglected that.

Are the included data files included because they are used in tests?  Should we zip them or something to de-text-ify them?

misc/gen_Opacity_Project_comparison_plot.jl
```diff
@@ -0,0 +1,84 @@
...
+
+    for elem in arr
+        if !(elem > 0)
+            continue
+        end
+        min_val = min(min_val, elem)
+        max_val = max(max_val, elem)
+    end
+    (min_val,max_val)
+end
```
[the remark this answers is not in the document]
mabruzzo: For an array `[-2,-1,0,1,2]`, `extrema` will return `(-2,2)`, while this function returns `(1,2)`

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,578 @@
...
+        _try_env_prefix_path("KORG_OP_CROSS_SECTION_DIR", string(species_name, ".txt"))
+    else
+        cross_sec_file
+    end
+
+    (data_dict, each_photo_ion_subtable(my_cross_sec_file))
+end
+
+
+@doc raw"""
```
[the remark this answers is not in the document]
mabruzzo: I'll look into this more, but I think it was related to avoid escaping every backslash in the latex equation

misc/gen_Opacity_Project_comparison_plot.jl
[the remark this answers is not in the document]
ajwheeler: Got it.  Not sure how I missed that.  Would something like
```julia
extrema(filter(x->x>=0, arr))
```
work?

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,578 @@
...
+        energy_ryd = abs(state_prop.excitation_potential_ryd)
+
+        # now iterate over Temperatures
+        for j in 1:length(T_vals)
+
+            # now compute the weighting of the current state under LTE
+            weight = inv_partition_func_val[j] * statistical_weight * exp(-energy_ryd * β_Ryd[j])
+
+            current_cross_section = (func.(photon_energies) .*
+                                     (1.0 .- exp.(-photon_energies.*β_Ryd[j]))
```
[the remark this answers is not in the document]
mabruzzo: (1 - boltzmann factor) is the correction for stimulated emission (under LTE)

src/continuum_opacity/opacity_metal.jl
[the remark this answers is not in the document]
ajwheeler: Don't worry about it if it works and you don't remember

test/continuum_opacity.jl
mabruzzo: Makes sense to me

src/continuum_opacity/opacity_metal.jl
mabruzzo: In any case, this data-structure is reflecting the format used within the Opacity Project (I'm not really planning to change the format...)

mabruzzo: I just confirmed that the ground state of neutral Helium has S = 0

ajwheeler (commented):
This isn't a complete review, but I have to leave work early today.  I will pick up on monday with the implementation of `weighted_bf_cross_section_TOPBase`.

test/utilities.jl
```diff
@@ -0,0 +1,128 @@
...
+- `atol`: The desired absolute tolerance
+- `err_msg`: An optional error message that can be printed when the arrays are not close
+- `error_location_fmt`: An optional keyword used to specify a function that creates a string 
+  describing the location in the arrays where values are unequal, given a `CartesianIndex` object.
+  This should not be specified in cases where actual and reference have different shapes.
+"""
+function assert_allclose(actual, reference; rtol = 1e-7, atol = 0.0, err_msg = nothing,
+                         error_location_fmt = nothing)
+    # make the initial check fast!
+    if !any( abs.(actual .- reference) .> (rtol .* abs.(reference) .+ atol))
```
ajwheeler, line 23:
```suggestion
    if all( abs.(actual .- reference) .<= (rtol .* abs.(reference) .+ atol))
```

test/utilities.jl
```diff
@@ -0,0 +1,128 @@
...
+        return true
+    end
+
+    common_layout = (size(actual) == size(reference)) && (ndims(actual) == ndims(reference))
```
ajwheeler, line 27: Isn't the second part redundant?

test/utilities.jl
```diff
@@ -0,0 +1,128 @@
...
+
+This is inspired by numpy's assert_allclose function. However, we've changed some default values.
+Additionally, in cases where actual and reference don't have equal values but do have the same 
+shape, the locations of the max errors are also printed.
+
+# Keywords
+- `rtol`: The desired relative tolerance
+- `atol`: The desired absolute tolerance
+- `err_msg`: An optional error message that can be printed when the arrays are not close
+- `error_location_fmt`: An optional keyword used to specify a function that creates a string
```
ajwheeler, line 16: Are you using these two arguments anywhere in the tests?  They seem a bit overengineered.  Am I missing an important use case?

test/utilities.jl
```diff
@@ -0,0 +1,128 @@
...
+    end
+
+    if common_layout
+        for index in CartesianIndices(size(actual))
+            func(actual[index], reference[index]; index = index)
+        end
+    else
+        func.(actual, reference)
+    end
+
```
ajwheeler, line 64: Does an implementation like this make sense?  Are you getting anything by avoiding broadcasting in the `common_layout` case?

```julia
diff = abs.(actual_val .- ref_val)
max_abs_ind = argmax(diff)
max_abs_val = diff[max_abs_diff]

relative_diff = diff ./ abs.(ref_val)
max_rel_ind = argmax.(relative_dif)
max_rel_val = max_rel_diff[max_rel_ind]
```

ajwheeler (changes requested):
Done!

test/utilities.jl
mabruzzo: Yeah, this is used by the `assert_allclose_grid` function for pretty printing (see below)

test/utilities.jl
mabruzzo: You're right. I was thinking of numpy's `size` and `ndim` functions when I wrote that

src/linelist.jl
```diff
@@ -157,12 +157,14 @@ const literals = (H_I=Species("H_I"), H_II=Species("H_II"), He_I=Species("He_I")
...
 function Base.show(io::IO, m::MIME"text/plain", s::Species)
     show(io, m, s.formula)
-    print(io, " ", roman_numerals[s.charge+1])
+    print(io, " ", get_roman_numerals(s))
 end
 
 ismolecule(s::Species) = ismolecule(s.formula)
 get_mass(s::Species) = get_mass(s.formula)
 get_atoms(s::Species) = get_atoms(s.formula)
+get_atomic_symbols(s::Species) = get_atomicsymbols(s.formula)
```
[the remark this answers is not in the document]
mabruzzo: oops. i should have deleted that line

src/continuum_opacity/opacity_metal.jl
```diff
@@ -0,0 +1,416 @@
...
+
+where the ``s`` subscript corresponds to a given species. In this equation:
+- ``g_{s,{\bf n}}`` is the statistical weight of state ``{\bf n}`` for species ``s``.
+- ``U_s`` is the partition function for species ``s``.
+- ``E_{s,{\bf n}}`` is the excitation potential of state ``{\bf n}`` for species ``s``. In other
+  words, its the energy difference between state ``{\bf n}`` and the ground state.
+- ``\sigma_{\lambda,{\rm bf}, s,{\bf n}}`` is the bound free cross-section (a.k.a. photo-ionization
+  cross section) of state ``{\bf n}`` for species ``s``.
+
+Under the assumption of LTE, the linear aborption coefficient for bound-free absorption is simply:
```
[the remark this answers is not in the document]
mabruzzo: Yeah, you're right

mabruzzo:
@ajwheeler All of your comments have been addressed. I just rebased off of master, once more. Pending any final comments this should be ready to merge.

test/utilities.jl
```diff
@@ -0,0 +1,108 @@
...
+            if length(vals) != size(actual)[ax]
+                throw(string("length(independent_vars[",ax,"][2]) should be ", length(vals)))
+            end
+
+            if isnothing(units)
+                push!(tmp, string(name, " = ", vals[index[ax]]))
+            else
+                push!(tmp, string(name, " = ", vals[index[ax]], " ", units))
+            end
+        end
```
ajwheeler, line 104: Can I make two requests?  The first is that you rewrite the loop into a map, which I think is easier to reasonable once you are used to it.  The second is that you rename `tmp` to something more meaningful.  Maybe `ax_strs`?  I'll leave it up to you whether you want to adopt the code golf demonstrated below to move the conditional inside the arguments of `string`.

```julia
ax_strs = map(1:ndims(actual)) do ax 

    ...

    string(name, " = ", vals[index[ax]], (isnothing(units) ? "" : " $units"))
end
```

test/utilities.jl
```diff
@@ -0,0 +1,108 @@
...
+
+    throw(ErrorException(join(lines, "\n")))
+    false
+end
+
+"""
+assert_allclose_grid(actual, reference, independent_vars; [rtol], [atol], [err_msg])
+
+This function is a convenience wrapper around assert_allclose that provides slightly nicer error 
+messages when actual and reference have the same shape.
```
ajwheeler, line 73: Feel free to reword, but I think it's worth explaining this a bit more.

```suggestion
This function is a convenience wrapper around assert_allclose that provides slightly nicer error 
messages based on the independent variables generating the grid.  It requires that actual and
reference have the same shape. 
```

src/linelist.jl
```diff
@@ -157,12 +157,13 @@ const literals = (H_I=Species("H_I"), H_II=Species("H_II"), He_I=Species("He_I")
...
 function Base.show(io::IO, m::MIME"text/plain", s::Species)
     show(io, m, s.formula)
-    print(io, " ", roman_numerals[s.charge+1])
+    print(io, " ", get_roman_numerals(s))
 end
 
 ismolecule(s::Species) = ismolecule(s.formula)
 get_mass(s::Species) = get_mass(s.formula)
 get_atoms(s::Species) = get_atoms(s.formula)
+get_roman_numerals(s::Species) = roman_numerals[s.charge+1]
```
ajwheeler, line 166: Sorry, I should have said this before, but I think it should be singular: `get_roman_numeral`.

PR #46, opened 2021-06-17
More broadening fixes

ajwheeler:
- don't approximate pressure broadening params for molecules
- skip pressure broadening of mols when doing synthesis
- filter higher ions from vald linelist during parse

I am accruing a lot of tech debt in the linelist parsing system.  Unfortunately, there's more changes sitting in #40 .  I will add a rewrite/refactor the the paper-blocking issues.

### DIFF
--- a/src/linelist.jl
+++ b/src/linelist.jl
@@ -131,6 +131,10 @@ Warner 1967.
 Used for atomic lines with no vdW and stark broadening info in the line list.
 """
 function approximate_gammas(wl, species, E_lower; ionization_energies=ionization_energies)
+    if ismolecule(species)
+        return 0.0,0.0
+    end
+
     ionization = split(species, '_')[2]
     Z = if ionization == "I"
         1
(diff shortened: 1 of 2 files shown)


src/linelist.jl
```diff
@@ -150,8 +154,14 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
     γstark = 0.77e-18 * nstar4_upper * wl^2
 
     Δrbar2 = (5/2) * RydbergH_eV^2 * Z^2 * (1/(χ - E_upper)^2 - 1/(χ - E_lower)^2)
-    #From R J Rutten's course notes, an equivalent form is also in Gray 2005
-    γvdW = 6.33 + 0.4log10(Δrbar2) + 0.3log10(10_000) + log10(k)
+    if χ < E_upper
+        println("Warning: for the $(species) line at $(Int(floor(wl*1e8))), the upper energy level"*
+                " exceeds the ionization energy (E_upper) > $(χ)). Using null broadening params.")
```
mabruzzo, line 159: One thought: would it be worth using  the warning syntax? (Note: I'm not actually sure that the following works)
```suggestion
        @warn ("For the $(species) line at $(Int(floor(wl*1e8))), the upper energy level exceeds"*
               " the ionization energy (E_upper) > $(χ)). Using null broadening params.")
```

mabruzzo (approved):
Looks good. I had one minor recommendation (feel free to ignore it)

src/linelist.jl
ajwheeler: I think I'm OK with this for now, especially since I'm requiring myself to rewrite this eventually anyway.  I'm going to ask Melissa about talking to Karen Lind (spectroscopist) about this issue.

mabruzzo: Got it. Merging

PR #48, opened 2021-06-22
fixed-width file parser

ajwheeler:
This is a basic parser.  I've already used it to parse turbospectrum line lists, but I haven't rolled that functionality into Korg yet.  It's still living in my notebook.  I haven't added any tests here, but I am going to rewrite all the lineline parsing code to use this, so it will be covered when it becomes critical.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -9,7 +9,7 @@ Applies a gaussian line spread function the the spectrum with flux vector `flux`
 vector `wls` with constant spectral resolution, R = λ/Δλ.
 
 This will have weird behavior if your wavelength grid is not locally linearly-spaced.
-It is intended to be run on a fine wavelength grid (``\\Delta\\lambda \\lesssim 0.05 \\AA``), then 
+It is intended to be run on a fine wavelength grid (``\\Delta\\lambda \\lesssim 0.05 \\AA``), then
 downsampled to the observational (or otherwise desired) grid.
 """
 function constant_R_LSF(flux::AbstractVector{F}, wls, R) where F <: Real
@@ -27,7 +27,7 @@ function constant_R_LSF(flux::AbstractVector{F}, wls, R) where F <: Real
 end
 
 """
-Rectify the spectrum with flux vector `flux` and wavelengths `wls` by dividing out a moving 
+Rectify the spectrum with flux vector `flux` and wavelengths `wls` by dividing out a moving
 `q`-quantile with `bandwidth`.
 """
 function rectify(flux::AbstractVector{F}, wls; bandwidth=50, q=0.95) where F <: Real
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -62,3 +62,50 @@ function vacuum_to_air(λ)
...
+        end
+        _parse_fwf_core(eachline(file), rowspec, lastrow-datarow+1)
+    end
+end
+function parse_fwf(lines::AbstractVector{String}, rowspec; datarow=1, lastrow=0)
+    lastrow = lastrow == 0 ? length(lines) : lastrow
+    _parse_fwf_core(lines[datarow:lastrow], rowspec)
+end
+
+function _parse_fwf_core(lines, rowspec, n=0)
```
mabruzzo, line 90: What is `n` doing here?

src/utils.jl
```diff
@@ -62,3 +62,50 @@ function vacuum_to_air(λ)
...
+    end
+end
+
+function _parse_item(line, itemspec)
+    if length(itemspec) == 4
+        r, t, n, f = itemspec
+    else
+        r, t, n = itemspec
+        f = identity
+    end
```
mabruzzo, line 103: It would probably make a sense to make sure that `itemspec` conforms to your expectations (and replacing it with a new value using the `identity` function) in `_parse_fwf_core` before executing `map` so that you don't execute this branch every single time you try to parse an entry.

src/utils.jl
```diff
@@ -62,3 +62,50 @@ function vacuum_to_air(λ)
...
+    if t == String
+        (n, f(line[r]))
+    elseif r[2] > length(line) || strip(line[r]) == ""
```
mabruzzo, line 106: Minor question: why are you checking to see if the second index of the range is longer than the line length? Checking if the first entry of the range is greater than the line length or the last entry is grater than the line length both seem like more robust checks.

src/utils.jl
```diff
@@ -62,3 +62,50 @@ function vacuum_to_air(λ)
...
 end
+
+"""
+    parse_fwf(source, rowspec; datarow, lastrow)
+
+Parse fixed-width files, returning a vector of `NamedTuples`.
+`source` can be either an `AbstractVector` of `Strings` or a filename.
+`rowspec` should be a vector of tuples each containing, in order
+- a range object corresponding to the column postion
+- the type.  Strings will pull pulled directly other types will be parsed with `parse`.
```
mabruzzo, line 73: typo:
```suggestion
- the type.  Strings will be pulled directly & other types will be parsed with `parse`.
```

mabruzzo (approved):

src/utils.jl
```diff
@@ -62,3 +62,50 @@ function vacuum_to_air(λ)
...
+end
+function parse_fwf(lines::AbstractVector{String}, rowspec; datarow=1, lastrow=0)
+    lastrow = lastrow == 0 ? length(lines) : lastrow
+    _parse_fwf_core(lines[datarow:lastrow], rowspec)
+end
+
+function _parse_fwf_core(lines, rowspec, n=0)
+    #use n
+    map(lines) do line
+        (; _parse_item.(line, rowspec)...)
```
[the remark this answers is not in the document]
ajwheeler: sorry this is "clever code"

src/utils.jl
ajwheeler: I should probably do better checking that `rowspec` is valid, but I'm not really worried about performance here.

PR #50, opened 2021-06-28
add missing factor of 2 to astrophysical flux integral

ajwheeler:
This add a missing factor of 2 to the final flux integral.  After doing this, our continuum in the optical closely matches MOOG's
image

In the IR, our continuum is about 10% higher that MOOG's, indicating that our opacities are too low. 
image


mabruzzo:
@ajwheeler I figured out the source of the discrepancy in the IR - it's my fault. I used a polynomial fit for H⁻ bound-free emission from Gray (2005), that truncates at 15000  Å. The original paper gives data out to 16300 Å. (At the time I assumed we would never go this far into the IR since other approximations break down). I'll try to fix this tonight.

The truncation in the following plot demonstrates that this explains the entirety of the discrepancy:
solar_comparison

ajwheeler:
Interesting, no rush. I can implement the changes if you want.

ajwheeler:
I’ll be very curious to see how this changes the Apogee spectra.

Issue #12, opened 2021-03-28
Outstanding Concerns about Continuum Opacities


mabruzzo:
This issue is being created as place to record some outstanding concerns about currently implemented continuum opacities. For the most part, these concerns shouldn't necessarily preclude their usage.

These concerns include:
 - The minimum wavelength at which we can currently compute He⁻ free-free opacity is 5063 Å.
 - I'm a little concerned about the bound-free opacities at really long wavelengths (say ≳ 4.45 microns), where our approximation of the summation with an integral becomes more aggressive than what Gray (2005) recommends.
-  I'm a little concerned about the differences in the bound-free opacities from MOOG. We use the same approach (described by Kurucz 1970), but with 2 differences and I'm concerned that they know something that we don't. These differences include: 
    1. MOOG seems to entirely ignore the opacity contributions of transitions originating from higher energy levels (which we use the integral approximation for). 
    2. MOOG uses a completely different table to compute the bound-free opacity for contributions originating from the ground state (this latter point is only relevant for wavelengths at or below 912 Å)
 - If we ever enforce a cuttoff energy level in our partition functions, it might be important to account for this in our bound-free opacities (this may actually have something to do with a difference highlighted in the previous bullet point).
 - I'm not extremely confident that our He⁻ free-free opacity is accurate. The uncertainties in the references values are comparable to the magnitude of the reference values. I've got part of a test written to help improve our confidence, but I'm not sure we'll ever be as confident about its correctness as I am about H⁻ bf, H⁻ ff, H I bf, or H I ff.
 - We don't currently have a test for the electron scattering opacity. It may not be necessary since the function is a single line featuring just a multiplication and a division. 
 - We don't have tests checking the accuracy of the He II ff opacities. However, the fact that these calculations wrap the same functions that are used for the H I bf and H I ff calculations definitely instills some confidence.

ajwheeler:
4.45 microns is a pretty big wavelength.  APOGEE, for example, extends to less that 2 microns.  It's great to have all this written down, though.

mabruzzo:
Yeah, I kind of assumed that was the case. It seems firmly in the mid-infrared regime

ajwheeler:
Summarizing current status of these things for posterity:

[...]

PR #51, opened 2021-06-29
Updated H⁻ bf opacity calculation

mabruzzo:
I altered `ContinuumOpacity.Hminus_bf` so that the cross-section is computed by directly interpolating data from Wishart (1979).

Previously, the cross-sections were computed using a polynomial from Gray (2005) that was fit to the tabulated data in Wishart (1979) while the tabulated data was used for verify the polynomial. Now, we use the Gray (2005) polynomial in our test to confirm that our transcription of the table is correct.

This change was primarily motivated by the fact directly using the table gives us opacities to wavelengths that are 1300 Å than before. 

While looking at [Mathisen (1984) ](https://inis.iaea.org/search/search.aspx?orig_q=RN:16033032), it clicked that the cross-section obviously goes to 0 at the ionization energy of H⁻.
- In other words,  there is no longer an upper wavelength limit where we don't know the H⁻ bound free opacity
- for reference, this zero-point is ~140 Å above the max wavelength in the table (in a part of the table where the wavelength spacing is 100 Å)
- my implementation for handling this is definitely not efficient. But I'm inclined to keep it until there is evidence that it is a performance bottleneck

I'm pretty confident that this addresses the IR solar spectrum discrepancy with `MOOG` that was described in PR #50. Here is the updated continuum spectrum:

solar_comparison

I wanted to note one thing about what `MOOG`'s handling of the H⁻ bf opacity:
- They follow the suggestion of [Mathisen (1984) ](https://inis.iaea.org/search/search.aspx?orig_q=RN:16033032), and use a different data source for wavelengths from 180 Å to 1640 Å. 
    - My inclination is not to do this, since it doesn't really get us a whole lot since our current source gets down to 1250 Å. I could add a note about this to Issue #12 in case there's ever any interest in sythesizing UV spectra.
    - Moreover, there are apparently some resonance effects below 1250 Å, and that makes the accuracy of the other source a little questionable.

### DIFF
--- a/src/continuum_opacity/opacity_H.jl
+++ b/src/continuum_opacity/opacity_H.jl
@@ -77,8 +77,8 @@ function _Hminus_bf_cross_section(λ, ion_energy_H⁻)
     Å_per_eV = 1e8 * (hplanck_eV * c_cgs)
     max_λ_ionize = Å_per_eV/ion_energy_H⁻
 
-    last_table_λ = _Hminus_bf_table[size(_Hminus_bf_table, 1), 1]
-    last_table_cross_section = _Hminus_bf_table[size(_Hminus_bf_table, 1), 2]
+    last_table_λ = _Hminus_bf_table[end, 1]
+    last_table_cross_section = _Hminus_bf_table[end, 2]
 
     @assert max_λ_ionize > last_table_λ
 
@@ -115,7 +115,7 @@ Compute the H⁻ bound-free opacity κ
 This function assumes that n(H⁻) ≪ n(H I) + n(H II). The number density of n(H⁻) should not be
 pre-computed (instead it's computed internally by this function).
 
-This function is adapted from equation 8.12 from Grey (2005) This equation gives the absorption
+This function is adapted from equation 8.12 from Grey (2005). This equation gives the absorption
 coefficient per H I atom (uncorrected by stimulated emission) is given by:
 
     ``8.316e-10 \\alpha_{bf}(H^-) P_e \\theta^{2.5} * 10^{{\rm ion_energy_H}^- \\theta} U(H⁻,T)/ U(H I,T)``
@@ -144,10 +144,10 @@ suggests that this data has better than 3% accuracy.
 function Hminus_bf(nH_I_div_partition::Real, ne::Real, ν::Real, ρ::Real, T::Real, 
                    ion_energy_H⁻::Real = _H⁻_ion_energy)
     λ = c_cgs*1e8/ν # in ångstroms
-    tmp = _Hminus_bf_cross_section(λ, ion_energy_H⁻) # in units of megabarn
+    cross_section = _Hminus_bf_cross_section(λ, ion_energy_H⁻) # in units of megabarn
     # convert from megabarn to cm² and include contributions from stimulated emission  1e-18
-    αbf_H⁻ = (1 - exp(-hplanck_cgs*ν/(kboltz_cgs*T))) * tmp * 1e-18
-    _ndens_Hminus(nH_I_div_partition, ne, T, _H⁻_ion_energy) * αbf_H⁻ / ρ
+    cross_section *= (1 - exp(-hplanck_cgs*ν/(kboltz_cgs*T))) * 1e-18
+    _ndens_Hminus(nH_I_div_partition, ne, T, _H⁻_ion_energy) * cross_section / ρ
 end
 
 
--- a/src/continuum_opacity/opacity_H.jl
+++ b/src/continuum_opacity/opacity_H.jl
@@ -61,17 +61,40 @@ const _Hminus_bf_table = SA[1250.0 5.431;   1500.0 6.512;   1750.0 7.918;   2000
                             15250.0 5.677;  15500.0 4.052;  15750.0 2.575;  16000.0 1.302;
                             16100.0 0.8697; 16200.0 0.4974; 16300.0 0.1989]
 
+const _Hminus_bf_cross_section_interp = LinearInterpolation(view(_Hminus_bf_table, :, 1),
+                                                            view(_Hminus_bf_table, :, 2),
+                                                            extrapolation_bc=Throw())
+
 """
-    _Hminus_bf_cross_section(λ)
+    _Hminus_bf_cross_section(λ, ion_energy_H⁻)
 
 Compute the H⁻ bound-free cross-section at a given wavelength (specified in Å). The cross-section 
 has units of megabarns per H⁻ particle and does NOT include a correction for stimulated emission.
 
 This function linearly interpolates the table provided in Wishart (1979).
 """
-const _Hminus_bf_cross_section = LinearInterpolation(view(_Hminus_bf_table, :, 1),
-                                                     view(_Hminus_bf_table, :, 2),
-                                                     extrapolation_bc=Throw())
+function _Hminus_bf_cross_section(λ, ion_energy_H⁻)
+    Å_per_eV = 1e8 * (hplanck_eV * c_cgs)
+    max_λ_ionize = Å_per_eV/ion_energy_H⁻
+
+    last_table_λ = _Hminus_bf_table[size(_Hminus_bf_table, 1), 1]
+    last_table_cross_section = _Hminus_bf_table[size(_Hminus_bf_table, 1), 2]
+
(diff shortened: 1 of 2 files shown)


src/continuum_opacity/opacity_H.jl
```diff
@@ -44,29 +44,56 @@ function _ndens_Hminus(nH_I_div_partition, ne, T, ion_energy = _H⁻_ion_energy)
...
-    λ = c_cgs*1e8/ν # in Angstroms
-    # we need to somehow factor out this bounds checking
-    if !(2250 <= λ <= 15000.0)
-        throw(DomainError(λ, "The wavelength must lie in the interval [2250 Å, 15000 Å]"))
+function _Hminus_bf_cross_section(λ, ion_energy_H⁻)
+    Å_per_eV = 1e8 * (hplanck_eV * c_cgs)
+    max_λ_ionize = Å_per_eV/ion_energy_H⁻
+
+    last_table_λ = _Hminus_bf_table[size(_Hminus_bf_table, 1), 1]
+    last_table_cross_section = _Hminus_bf_table[size(_Hminus_bf_table, 1), 2]
```
ajwheeler, line 81:
```suggestion
    last_table_λ = _Hminus_bf_table[end, 1]
    last_table_cross_section = _Hminus_bf_table[end, 2]
```

src/continuum_opacity/opacity_H.jl
```diff
@@ -44,29 +44,56 @@ function _ndens_Hminus(nH_I_div_partition, ne, T, ion_energy = _H⁻_ion_energy)
...
+
+    @assert max_λ_ionize > last_table_λ
+
+    if λ < _Hminus_bf_table[1,1]
+        throw(DomainError(λ, "$(_Hminus_bf_table[1,1]) Å must not exceed the wavelength"))
```
ajwheeler, line 86: Is this `>` the right way around?

ajwheeler: Or the word "exceed" wrong?

src/continuum_opacity/opacity_H.jl
```diff
@@ -88,11 +115,14 @@ Compute the H⁻ bound-free opacity κ
 This function assumes that n(H⁻) ≪ n(H I) + n(H II). The number density of n(H⁻) should not be
 pre-computed (instead it's computed internally by this function).
 
-This function is adapted from equation 8.12 from Grey (2005). This equation gives the absorption
+This function is adapted from equation 8.12 from Grey (2005) This equation gives the absorption
```
ajwheeler, line 118:
```suggestion
This function is adapted from equation 8.12 from Grey (2005). This equation gives the absorption
```

src/continuum_opacity/opacity_H.jl
```diff
@@ -101,26 +131,23 @@ Eqn 8.12 of Grey (2005) implicitly assumes that (U(H⁻,T)/ U(H I,T)) is always
...
 function Hminus_bf(nH_I_div_partition::Real, ne::Real, ν::Real, ρ::Real, T::Real, 
                    ion_energy_H⁻::Real = 0.7552)
-    αbf_H⁻ = _Hminus_bf_cross_section(ν) # does not include contributions from stimulated emission
-    stimulated_emission_correction = (1 - exp(-hplanck_cgs*ν/(kboltz_cgs*T)))
-    n_H⁻ = _ndens_Hminus(nH_I_div_partition, ne, T, _H⁻_ion_energy)
-    αbf_H⁻ * n_H⁻ * stimulated_emission_correction / ρ
+    λ = c_cgs*1e8/ν # in ångstroms
+    tmp = _Hminus_bf_cross_section(λ, ion_energy_H⁻) # in units of megabarn
+    # convert from megabarn to cm² and include contributions from stimulated emission  1e-18
+    αbf_H⁻ = (1 - exp(-hplanck_cgs*ν/(kboltz_cgs*T))) * tmp * 1e-18
```
ajwheeler, line 149: Do you have a source for the `1e18`?

src/continuum_opacity/opacity_H.jl
```diff
@@ -101,26 +131,23 @@ Eqn 8.12 of Grey (2005) implicitly assumes that (U(H⁻,T)/ U(H I,T)) is always
...
+    _ndens_Hminus(nH_I_div_partition, ne, T, _H⁻_ion_energy) * αbf_H⁻ / ρ
```
ajwheeler, line 150: `αbf_H⁻` isn't really an alpha value is it?

```suggestion
    cross_section = _Hminus_bf_cross_section(λ, ion_energy_H⁻) # in units of megabarn
    # convert from megabarn to cm² and include contributions from stimulated emission  1e-18
    cross_section *= (1 - exp(-hplanck_cgs*ν/(kboltz_cgs*T))) * 1e-18
    _ndens_Hminus(nH_I_div_partition, ne, T, _H⁻_ion_energy) * cross_section / ρ
```

ajwheeler (changes requested):

src/continuum_opacity/opacity_H.jl
mabruzzo: There are 1e18 megabarn in 1 cm². I'll clean up the code to be more clear

ajwheeler:
Ah, that was a thinko on my part. I don’t think you need to
change anything.

src/continuum_opacity/opacity_H.jl
mabruzzo: This is the Gray (2005) notation. They use α to refer to the cross-section (unlike Rybicki & Lightman who I think use σ?). I'll definitely make this change, but it's something to keep in mind documentation-wise.

src/continuum_opacity/opacity_H.jl
ajwheeler: This is fine as is.  It was a thinko on my part.

src/continuum_opacity/opacity_H.jl
ajwheeler: Ah yes, sorry I keep forgetting that.

src/continuum_opacity/opacity_H.jl
mabruzzo: This was technically correct as written, but it definitely looked weird (I basically flipped the inequality). I changed it so that it makes more sense.

mabruzzo:
Minor heads up: I think I introduced a new test set to improve test coverage. I think I pushed that just after you finished your review, so you may have missed it. I think I addressed all of your other concerns.

ajwheeler:
image

src/continuum_opacity/opacity_H.jl
ajwheeler:
```suggestion
        throw(DomainError(λ, "λ must be ≥ $(_Hminus_bf_table[1,1]) Å"))
```

ajwheeler: Ooooooh, this took me way longer than it should have.

ajwheeler (changes requested):

src/continuum_opacity/opacity_H.jl
mabruzzo: I'll apply this later (I can't apply it easily now since you didn't make the suggestion using the latest changeset)

src/continuum_opacity/opacity_H.jl
```diff
@@ -44,29 +44,56 @@ function _ndens_Hminus(nH_I_div_partition, ne, T, ion_energy = _H⁻_ion_energy)
...
+    Å_per_eV = 1e8 * (hplanck_eV * c_cgs)
+    max_λ_ionize = Å_per_eV/ion_energy_H⁻
+
+    last_table_λ = _Hminus_bf_table[end, 1]
+    last_table_cross_section = _Hminus_bf_table[end, 2]
+
+    @assert max_λ_ionize > last_table_λ
+
+    if λ < _Hminus_bf_table[1,1]
+        throw(DomainError(λ, "wavelength must be greater or equal to $(_Hminus_bf_table[1,1]) Å."))
```
ajwheeler, line 86:
```suggestion
        throw(DomainError(λ, "λ must be ≥ $(_Hminus_bf_table[1,1]) Å"))
```

ajwheeler: I think I did it right this time.

PR #54, opened 2021-07-23
H lines

ajwheeler:
- ~resonant broadening of Balmer series~
- ~what's the most appropriate convolution approximation?~ _I've avoided this problem by only including self-broadening for Halpha-Hgamma and only including stark broadening for other H lines.  This is not perfect, but we aren't getting these lines completely right without NLTE anyway._
- ~set versions for new dependencies~


src/line_opacity.jl
```diff
@@ -76,12 +74,116 @@ function line_absorption(linelist, λs, temp, nₑ, n_densities::Dict, partition
...
+             log_gf = HDF5.read_attribute(fid[transition], "log_gf"),
+             λ0 = LinearInterpolation((temps, nes), read(fid[transition], "lambda0") * 1e-8) 
+            )
+        end
+    end
+end
+
+
+#used in hydrogen_line_absorption
+_zero2epsilon(x) = x == 0 ? floatmin() : x
```
mabruzzo, line 117:
```suggestion
_zero2epsilon(x) = x + (x == 0) * floatmin()
```
This is just a micro-optimization to explicitly avoid branches. Feel free to ignore it.

mabruzzo (approved):
This looks good to me. I just had a couple of minor questions.

Out of curiosity, how does this affect speed and agreement with turbospec?

ajwheeler:
I haven't compared to turbospec yet (hopefully later today), but it's pretty fast – 0.02s (per atmospheric layer) for all H lines across "all" (3000-17000 A) wavelengths with a 0.01 A spacing.  I haven't profiled so not sure how it compares to the continuum opacities, but it's definitely closer to those than it is to general line opacities.

mabruzzo:
Minor aside, while it's still fresh in your mind how to use the Barklem code, would it potentially be useful to use that code to generate some sample line absorption profiles and use those as unit tests? (Feel free to ignore this idea)

ajwheeler:
I was waffling quite a bit about this.  Let's talk about it in person or
zoom sometime.

test/runtests.jl
```diff
@@ -241,6 +240,24 @@ end
             @test 0.99 < sum(ϕ .* Δ)/amplitude < 1
         end
     end
+
+    @testset "hydrogen stark profiles" begin
```
mabruzzo, line 244: Minor suggestion: I would just add a brief comment clarifying where the test data comes from.

Maybe you could do something like the following?
```suggestion
    @testset "hydrogen stark profiles" begin
        # This test data was generated with Korg.hydrogen_line_absorption shortly
        # after writing the function. This data is consistent with the results
        # produced by the Fortran code distributed with Barklem+ 2000
```

mabruzzo (approved):
Looks good to go! I had one minor suggestion.

test/runtests.jl
ajwheeler: It's actually Stehle + 1999.  I've committed this comment but with the correct paper substituted.  (Barklem 2000 is the self-broadening profiles, for which we just use the standard ABO machinery).

Issue #10, opened 2021-03-20
Gold Standard Testing


mabruzzo:
This issue is for discussing the design requirements for "gold standard testing" (since this is not necessarily obvious). I wanted to flesh out my own ideas and hear your ideas. As I write this out, I'm becoming les

**High Level Description:** The basic idea is to perform calculations under a standardized set of conditions and then cache the results (which serve as our "gold standard"). Then, after making changes we perform the same calculations and compare our result 

**Objectives:** I would argue that there are 3 main objectives for this tool
1. Provide confidence that changes to one part of the code don't produce unexpected altered results in another part of the code.
2. Provide confidence that our intermediate calculations (e.g. continuum opacity) give an accurate answer. Unfortunately there is no "standardized answer set" for the intermediate calculations.  In a sense the "gold standard" acts as our bootstrapped answer set.
3. (Optional) Possibly reuse the code both for unit tests and integration/end-to-end tests.

**High-Level Design:** To be useful, I think our tools need to consist of 3 pieces
1. We obviously need to run the test
2. We need to be able to generate new gold standard answers
3. We need to be able to easily visualize the difference between new results and the gold standard if there is disagreement. This might sound unnecessary, but I think this is crucial. I'm not sure what the best way to accomplish this is. (Ordinarily, I would recommend making a standalone command-line script. I'm less sure about what to do since this is Julia).

**Other considerations:**
- What features are we using this approach to test?
- Should this work through the traditional Julia unit-testing code? I'm starting to think that the answer is "no"
- How do we want to serialize results?
  - Should we package results for tests using similar functionallity together or keep one test per file?
  - Should we use HDF5? If we package things in individual files, JSON might be better? If we want to use this for full integration tests, what is the output format of our synthetic spectra going to be?
  - How much meta-data do we want to store in the result?
- We need to make sure that there is a straight-forward procedure for introducing new tests

<details>
  <summary> Comparison to other testing approaches </summary>

  "Gold standard" testing isn't the only tool at our disposal for verifying that our intermediate calculations give the correct result. Other types of tests include: 
1. Comparing our synthetic spectra to the spectra for stars with known properties or the synthetic spectra produced by MOOG/Turbospectrum
2. There are some figures showing intermediate results of our calculations. For example, this is currently what we are doing with our continuum opacity calculations.
3. We could imagine extracting equivalent sections of code from MOOG/Turbospectrum and produce tables of expected results.

[...]

PR #56, opened 2021-08-02
automating scientific sanity checks

ajwheeler:
This is the beginnings of an implementation of "gold standard testing" (#10) , which I/we are hesitantly renaming "scientific sanity checking".  I've created a sanity checking directory with it's own julia environment, and written a script to save a few regions of a generated solar spectrum in a HDF5 file.

From the `sanity_checks` directory, you can run `julia --project generate_data.jl out.h5` to run the script.

**edit:**  Also, since we will likely collaborate on this PR, I will not do any rebasing nonsense.


ajwheeler:
Closing because Andy is working on this now.

PR #57, opened 2021-08-09
Correct the unit problem with K (use cgs, not mks)

ajwheeler:
Also,
	•	use the no-molucule solution as a starting point fo moleq calculations
	•	consolidate loading statmech quantities into a single file
	•	improve documentation in a few places
	•	correctly parse numerical molecule code which have had missing zeros stripped.
	•	save and return the number density of each species at each layer in the atmosphere in `synthesize` (this API will be refined later)


src/read_statmech_quantities.jl
```diff
@@ -21,33 +51,27 @@ end
...
-at an input temperature.
+Constructs a Dict holding tables containing partition function or equilibrium constant values across
+temperature.
 
-In more detail, the partition function is linearly interpolated from values tabulated from Barklem 
-& Collet (2016). The partition functions can only be evaluated at temperatures between 1e-5 K and 
-1e4 K. The tabulated values were NOT computed with a cutoff principle quantum number. This can lead
-to a couple percent error for some neutral species in solar type stars (I think the problem becomes
-worse in hotter stars). See the paper for more details.
+The optionsl argument, `transform` is applied to each value. It is used to do unit conversions, etc.
```
mabruzzo, line 71:
```suggestion
The optional argument, `transform` is applied to each value. It is used to perform unit conversions,
when loading equilibrium constants
```
I fixed a typo and added a clause to clarify the usage of the keyword (I was originally very confused since partition functions don't have units)

mabruzzo (approved):
I had one minor suggestion. Otherwise, it looks great!

PR #58, opened 2021-08-18
linelist rewrite and moleq refactor

ajwheeler:
- [x] new `Species`, `Baryon` (?) types
- [x] overhaul VALD linelist parsing to use `CSV.jl`
- [x] add tests for all varieties of vald linelist
- [ ] ~~turbospectrum linelist format~~ _Maybe in the future.  I need more examples to work from and the APOGEE linelist has fully capitalized species names (e.g. "FEH") which are annoying to parse_.
- [ ] ~~automatic isotope scaling~~  _This is too annoying to parse for vald linelists, and simple to do manually.  I don't think it's a deal breaker._
- [x] perhaps predefined constants `H_I` and `H_II` (and possibly the analogous for helium) would be useful.
- [ ] ~~use fwf parser or CSV for moog and kurucz linelists~~ _Not worth it.  They are very simple as is.  I can use the fwf parser if I need more functionality later_
- [x] reconsider `Line` constructor situation (maybe pass missing for missing values, handle broadening approximations in the constructor)
- [x] make sure this stupid `Baryon` implementation isn't a performance regression.


ajwheeler:
@mabruzzo feel free to review as much or little as you like, or just ask me questions about what I did.

You will be interested to know that I created package global consts, `H_I`, `He_II`, etc. (just H and He) to avoid expensive parsing or long constructor calls (e.g. `Species(Baryon(["H"]), 0)` is fast but clunky and `Species("H I")` is simple-ish but slow).  Anyway with those in use, performance is the same (to within measurement error) as before.

Also, I really don't like the name `Baryon`, so if we can think of something else that would be great.

src/linelist.jl
```diff
@@ -1,132 +1,229 @@
...
-    count = 0
-    for c in species
-        if c == '_'
-            break
-        else
-            count += isdigit(c) + isuppercase(c)
-        end
-    end
-    count > 1
+function mass(b::Formula)
```
mabruzzo, line 105: Yeah, it would be great if we could come with a better name for this.

test/opacity_comparison_funcs.jl
```diff
@@ -216,27 +216,17 @@ function Heminus_ff_coefficient(λ, T, Pₑ)
...
+    # We use the ionization energy tabulated in table D.1 (χ(He I) = 24.587). 
     # Table D.2 of Gray (2005) records the partion function of He II to be
     # as 10^0.301, the closest value at the table's precision to 2.0
-    Us = [T -> 1.0,      # Table D.2 of Gray (2005) records the partion function of He II to be
-          T -> 10^0.301, # as 10^0.301, the closest value at the table's precision to 2.0
-          T -> NaN]
+    # Because this neglects He II, we don't use Korg.saha_ion_weights.
+    wII = (2.0/ne * (10^-0.301) * Korg.translational_U(Korg.electron_mass_cgs, T) 
+           * exp(-24.587/(Korg.kboltz_eV*T)))
+    nHe_I = nHe * 1/(1+wII)
```
mabruzzo, line 226:
```suggestion
    UI, UII = (1.0, 10^-0.301)
    # Because this neglects He II, we don't use Korg.saha_ion_weights.
    wII = (2.0/ne * (UII/UI) * Korg.translational_U(Korg.electron_mass_cgs, T) 
           * exp(-24.587/(Korg.kboltz_eV*T)))
    nHe_I = nHe * 1/(1+wII)
```

I know that this is an extra line, but I think it improves readability to include `UI = 1.0`. (The alternative is including a comment reminding the reader that UI = 1.0, because things looked wrong)

test/opacity_comparison_funcs.jl
```diff
@@ -216,27 +216,17 @@ function Heminus_ff_coefficient(λ, T, Pₑ)
...
-    wII, wIII = Korg.saha_ion_weights(T, ne, χs, Us)
-    nHe_I = nHe / (1 + wII) #Gray ignores He III
-
-    # may want to include the temperature dependence of the partition function
-    partition_func = 2.0
     ν = (Korg.c_cgs*1e8)/λ
-
-
-    opacity = Korg.ContinuumOpacity.Heminus_ff(nHe_I/Us[1](T), ne, ν, ρ, T)
+    opacity = Korg.ContinuumOpacity.Heminus_ff(nHe_I, ne, ν, ρ, T)
```
mabruzzo, line 229:
```suggestion
    opacity = Korg.ContinuumOpacity.Heminus_ff(nHe_I/UI, ne, ν, ρ, T)
```

src/synthesize.jl
```diff
@@ -154,20 +152,20 @@ function total_continuum_opacity(νs::Vector{F}, T::F, nₑ::F, ρ::F, number_de
...
     #He continuum opacities
-    κ += ContinuumOpacity.He_II_bf.(number_densities["He_II"]/partition_funcs["He_II"](T), νs, ρ, T)
-    #κ += ContinuumOpacity.He_II_ff.(number_densities["He_III"], nₑ, νs, ρ, T)
+    κ += ContinuumOpacity.He_II_bf.(number_densities[H_II] / partition_funcs[H_II](T), νs, ρ, T)
+    #κ += ContinuumOpacity.He_II_ff.(number_densities[He_III], nₑ, νs, ρ, T)
     # ContinuumOpacity.Heminus_ff is only valid for λ ≥ 5063 Å
-    #κ += ContinuumOpacity.Heminus_ff.(number_densities["He_I"]/partition_funcs["He_I"](T),
-    #                                  nₑ, νs, ρ, T)
+    #κ += ContinuumOpacity.Heminus_ff.(number_densities[He_I] / 
+    #        partition_funcs[He_I](T), nₑ, νs, ρ, T)
```
mabruzzo, line 168: Purely stylistic question: why did you move the divisor to the second row?  Just curious

src/linelist.jl
```diff
@@ -1,132 +1,229 @@
...
-    el1, el2        
+    Species(formula, charge)
+end
+
+#useful constants
+const H_I = Species("H_I")
+const H_II = Species("H_II")
+const He_I = Species("He_I")
+const He_II = Species("He_II")
+const He_III = Species("He_III")
```
mabruzzo, line 151: As we previously discussed, I think it would be a lot better if these could be placed in a submodule of some kind so that these constants are just floating in the global namespace.

src/linelist.jl
```diff
@@ -1,132 +1,229 @@
...
-    parse_species_code(code)
-
-Parse the "species code" as it is often specified in linelists and return a the "astronomy" 
-notation. 01.00 → "H_I", 02.01 → "He_II", 02.1000 → "He_II", 0608 → "CO_I", etc.  
+Represents an atom or molecule, irespective of its charge.
 """
-function parse_species_code(code::AbstractString)
-    toks = split(code, '.')
+struct Formula
+    atoms::Vector{String}
```
mabruzzo, line 8: As we previously discussed, I definitely think it would be useful to make the ``atoms`` field a static vector of 3 or 4 `UInt8`/`Int8` values. But I understand if you want to punt on that (for now)

mabruzzo (approved):

src/linelist.jl
ajwheeler: Thoughts on `get_mass`?  I don't love it but I can't think of anything better?

test/opacity_comparison_funcs.jl
ajwheeler: 😨

src/synthesize.jl
ajwheeler: Honestly no idea.

src/linelist.jl
ajwheeler: I think a submodule consisting only of `Formula`, `Species`, these, and well-known (i.e. not passable by the user) atomic data would make sense.  It could be called `Chemistry` or `Particles`?

src/linelist.jl
ajwheeler: Yeah I'm going to add this to the TODO issue and I'd love to tackle it before 1.0, but I'd like to merge this PR with the current implementation.

src/linelist.jl
mabruzzo: Yeah, that or `particle_mass` would work. 

Alternatively, I think the short name would probably be fine if we explicitly placed this in some submodule (maybe a `Species` submodule?)

ajwheeler: > Alternatively, I think the short name would probably be fine if we explicitly placed this in some submodule (maybe a Species submodule?)

True!  I guess that's kinda what they are for...

src/linelist.jl
mabruzzo: Makes sense to me

src/linelist.jl
mabruzzo: Sounds like a plan. 

If there's a pre-1.0 crunch , I would suggest that we at least convert from a `Vector{String}` to a `StaticVector{String}` (it bothers me a little that these datatypes are technically mutable)

src/linelist.jl
ajwheeler: I started to do this, but I'm not convinced anymore.  I think the clunkiness of julia's include/import system relative to python combined with how multiple dispatch makes the code much harder to follow.  For the continuum opacities, which are used in exactly one place, it nicely separates functionality and namespaces, but this code is so integrated with everything else, I don't think it makes clearer to have a bunch of qualified imports.  I'm still open to doing this in the future, but I think it would make more sense to refactor the whole package into submodules rather than doing it piecemeal.

ajwheeler: As an alternative solution to this problem, what about placing these in a (`const`) named tuple?

ajwheeler:
I've rewritten to use atomic numbers instead of symbols throughout and use a StaticArray for the Formula internals.

I also couldn't resist refactoring the moleq code a bit.  Sorry for making an over-broad PR.

These changes cut moleq time to roughly 0.1 seconds, and cut about 10-20% off synthesis time for the APOGEE test case we've been using.

ajwheeler:
This is ready for final review.

src/statmech.jl
```diff
@@ -13,31 +13,25 @@ arguments:
 - `ionization_energies` is a Dict mapping elements to their first three ionization energies
```
mabruzzo, line 13:
```suggestion
- `ionization_energies` is a Dict or Vector mapping elements' atomic numbers to their first three
  ionization energies
```

src/statmech.jl
```diff
@@ -13,31 +13,25 @@ arguments:
...
 - `partition_funcs` is a Dict mapping species to their partition functions
 """
-function saha_ion_weights(T, nₑ, atom::String, ionization_energies::Dict, partition_funcs::Dict)
+function saha_ion_weights(T, nₑ, atom, ionization_energies, partition_funcs::Dict)
```
mabruzzo, line 16: Github isn't letting me directly suggest this (since it hasn't been modified), but could you modify update the definition of the `atom` argument. Maybe you can make it say:

```
- atom, the atomic number or Formula of the element
```

(I would definitely err on the side of being too explicit here, because I think that the code can be a little hard to follow)

src/statmech.jl
```diff
@@ -13,31 +13,25 @@ arguments:
...
+    UI = partition_funcs[Species(atom, 0)](T)
+    UII = partition_funcs[Species(atom, 1)](T)
 
     k = kboltz_eV
     transU = translational_U(electron_mass_cgs, T)
     
-    wII =  (2.0/nₑ * (UII/UI) * transU * exp(-χI/k/T))
-    wIII = if atom == "H"
+    wII =  2.0/nₑ * (UII/UI) * transU * exp(-χI/(k*T))
+    wIII = if atom == Formula(1)
```
mabruzzo, line 26: Feel free to ignore this if you want. But, I think this may make things more clear
```suggestion
    wIII = if atom == Formula(1) # hydrogen
```

mabruzzo (changes requested):
A couple of minor suggestions, but otherwise, I think it's good to. 

I'm personally not a huge fan of the transition to using vectors for atomic masses and ionization energies instead of dicts (I think it really hinders readibility). But, I've made my peace with the change.

As an aside, a useful micro-optimization for the future would be to change the `charge` field of the `Species` struct to be of type `Intt8` (doing this would cut the size of `Species` from 11 bytes down to 4 bytes). But I definitely wouldn't worry about this right now.

ajwheeler:
>I'm personally not a huge fan of the transition to using vectors for atomic masses and ionization energies instead of dicts (I think it really hinders readibility). But, I've made my peace with the change.
For me, this comes from a desire to represent atoms with their atomic numbers everywhere (except in the arguments to the user-facing top-level `synthesize`).  Since atoms are represented with atomic numbers inside `Formula`, it would be weird indirection to look up their symbols to plug into the mass and ionization dicts.

src/statmech.jl
ajwheeler: It can't actually be a `Formula`, right?

ajwheeler:
>As an aside, a useful micro-optimization for the future would be to change the charge field of the Species struct to be of type Intt8 (doing this would cut the size of Species from 11 bytes down to 4 bytes). But I definitely wouldn't worry about this right now.

Is there any downside?  Might as well...

ajwheeler:
OK, I tried it and synthesis got slightly slower.  Will add to the TODO issue to investigate in future.

PR #62, opened 2021-08-31
Voigt speedup

ajwheeler:


mabruzzo:
When you get a chance, can you reset this PR (so that it doesn't include changes from a different PR)

PR #72, opened 2021-09-13
add H and He Rayleigh scattering

ajwheeler:
This adds Rayleigh scattering from H and He.  I have not yet added any tests, nor have I called into this anywhere.

(Note that I initially accidentally committed this to main and pushed 🤦 .  I `git reverted` that and `git reverted` the revert commit in this PR.)

These cross section are very small.  Small enough that I suspect they are well-within the errors of more dominant absorption sources. 

The [Colgan+ paper](https://iopscience.iop.org/article/10.3847/0004-637X/817/2/116/pdf) doesn't include a formulation for H2.  MOOG assigned it a cross section similar to, and thus an alpha value much smaller than that for H.


ajwheeler:
I've now added Rayleigh scattering from H2 (using the formula from http://articles.adsabs.harvard.edu/pdf/1962ApJ...136..690D).  This appears to still be standard.  See for instance, [petiteRADTRANS](https://ui.adsabs.harvard.edu/abs/2019A%26A...627A..67M/abstract) (for transmission spectroscopy of exoplanet atmospheres).  It is also what MOOG and Turbospectrum use. 

It's not clear to me how to test this.

ajwheeler:
**update: fixing bugs, don't review**

ajwheeler:
Done.
image

mabruzzo (approved):
I've got a single comment. But other than that, I think it's ready to merge.

ajwheeler:
I've addressed your comments.  Does it make sense to hard code the wavelength bound rather than use your wrapper function?

PR #73, opened 2021-09-19
WIP: Transition Continuum Opacities to Continuum Absorption Coefficients

mabruzzo:
I'm more than half-way there... I just need to handle the boundary conditions.


ajwheeler:
TODO: CamelCase

ajwheeler:
OK, I think this overall architecture makes sense, with the caveat that I think the contents of the `bounds_checking.jl` should be simplified a lot.  I like the convention of using underscores for the the non-bounds-checks functions.

- As we discussed, I think you should `@assert` that lambda/nu arrays are sorted.  That should simplify things a lot.
- Like you said, I think the string formatted intervals are neat, but unnecessarily complicated.  I would prefer not to maintain that.
- I don't think you should support bounds in lambda.  Can't these be converted into nu-bounds at implementation time?  If you really think it's worth automating, you could provide code to do it but call it when defining the interval.  (I'm skeptical that this is worth the extra code, though.)
- Have you considered arguments for multiple bound types in `bounds_checked_absorption`, e.g. `nu_bounds`, and `nu_over_T_bounds`, rather than using `photon_bound_type`?  I find this to be a bit finicky and it can fail at run time. ( I realized that doesn't matter in the JIT regime, but hopefully Julia's static analysis situation improves eventually.)  This also has the advantage that you can support bounds in both nu and nu/T for the same function. 
- I think `wrapped_func` should either add to an alpha array or return an array.  I definitely support the python "one true way" ideology here.  Having the alpha array as the first argument (and naming the functions with a `!`) would also be more idiomatic.

Hopefully, this isn't too frustrating.  Let me know what you think.

mabruzzo:
>     * As we discussed, I think you should `@assert` that lambda/nu arrays are sorted.  That should simplify things a lot.
> 
>     * Like you said, I think the string formatted intervals are neat, but unnecessarily complicated.  I would prefer not to maintain that.
> 
>     * I don't think you should support bounds in lambda.  Can't these be converted into nu-bounds at implementation time?  If you really think it's worth automating, you could provide code to do it but call it when defining the interval.  (I'm skeptical that this is worth the extra code, though.)
This above all largely sounds fine to me. We can iterate more on the lambda bounds, I've got some thoughts here (but I don't think they're incompatible with what you're saying here).

>     * Have you considered arguments for multiple bound types in `bounds_checked_absorption`, e.g. `nu_bounds`, and `nu_over_T_bounds`, rather than using `photon_bound_type`?  I find this to be a bit finicky and it can fail at run time. ( I realized that doesn't matter in the JIT regime, but hopefully Julia's static analysis situation improves eventually.)  This also has the advantage that you can support bounds in both nu and nu/T for the same function.

I had already given this some thought to doing exactly this, but I just hadn't gotten there yet.

>     * I think `wrapped_func` should either add to an alpha array or return an array.  I definitely support the python "one true way" ideology here.  Having the alpha array as the first argument (and naming the functions with a `!`) would also be more idiomatic.

This is the only thing that I wanted to lightly push back on. What I'm doing is actually very much inspired by numpy. Would it be more tolerable, if the function overwrote the contents of the `out` keyword? If you feel strongly about this, I'll either:
- always have wrapper_func allocated it's own memory
- or think about integrating all of the bounds checking into `total_continuum_absorption`

ajwheeler:
Let's talk IRL about that last point, I think we may be speaking past each other.  If you feel strongly I won't die on this hill.

mabruzzo:
@ajwheeler This is ready for review.

After we merge in #42, I'll reset the diff.

docs/src/API.md
```diff
@@ -39,17 +39,35 @@ Korg.voigt
 ## Continuum absorption
 These function can be used to directly compute continuum opacities.
 
+##### Continuum Absorption Kwargs
+
+Many of the functions documented in this section, support the following, common set of keyword
+arguments:
```
ajwheeler, line 45:
```suggestion
All bound-free and free-free opacity functions support these two keyword arguments.
```

test/absorption_comparison_funcs.jl
```diff
@@ -106,7 +107,22 @@ function _semi_realisitic_dens(ne::F, fion::F= 0.02, HydrogenMassFrac::F= 0.76)
...
 end
 
-# Now actually define the functions that compute the opacities in the form comparable with Gray05
+# Gray05 uses some confusing terminology and notation.
+# - he defines the "atomic absorption coefficient" as the area per absorber and represent it with
+#   the variable α. For reference, Rybicki & Lightman call this same quantity the cross-section and
+#   represent it with the variable σ_ν.
+# - he defines the "continuous absorption coefficient per neutral hydrogen atom" and denotes it
+#   with κ. Rybicki & Lightman don't explicitly define this quantity, but it's compatible with
+#   their nomenclature. R&L would express this quantity as α_ν/nH_I (in their notation)
```
ajwheeler, line 116: This quantity is actually the same as the cross section!

test/absorption_comparison_funcs.jl
```diff
@@ -106,7 +107,22 @@ function _semi_realisitic_dens(ne::F, fion::F= 0.02, HydrogenMassFrac::F= 0.76)
...
+# - he uses κ_ν to denote the mass opacity (just like Rybicki & Lightman)
+#
+# Now we actually define the functions that compute the opacities in the form comparable with
+# Gray05. These functons return the "continuous absorption coefficient per neutral hydrogen atom"
+# divided by Pₑ, the partial electron pressure.
+#
+# For the sake of clarity, we label Rybicki & Lightman's α_ν = κ_ν*ρ as the "linear absorption
+# coefficient" within these functions (note: throughout the rest of the codebase, we interchangably
+# use the simpler term, "absorption coefficient", to refer to this same quantity)
```
ajwheeler, line 125: How would you feel if I lifted this and placed it in the dev docs at a later point?

ajwheeler: (and removed it/referenced the dev docs here)

ajwheeler (commented):
Looking good.  

- As we discussed already, I think you should go from `extrapolation_bc` to a binary flag for whether or not to throw on oob.
- Do you need to rebase/merge to incorporate Rayleigh scattering?
- synthesize doesn't yet pass a range to total_continuum_absorption.  I'm happy to do this because it overlaps with what I want to do with the rest of the code, just bringing it up.
- Finally, I think you should consider moving the Interval stuff to utils.  The reason is that it would be very useful when converting wavelength _ranges_ to frequency ranges.

src/continuum_absorption/bounds_checking.jl
```diff
@@ -0,0 +1,180 @@
...
+function contained_slice(vals::AbstractVector, interval::Interval)
+    if first(vals) < last(vals)
+        first_ind = searchsortedfirst(vals, interval.lower)
+        last_ind = searchsortedlast(vals, interval.upper)
+    else
+        first_ind = searchsortedlast(vals, interval.upper; rev = true) + 1
+        last_ind = searchsortedfirst(vals, interval.lower; rev = true) - 1
+    end
+    first_ind:last_ind
+end
```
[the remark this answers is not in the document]
ajwheeler: It's arguable that it wouldn't, but I was think about something along these lines
```julia
lb = searchsortedfirst(vals, max(ν_bound.lower, ν_div_T_bound.lower * T))
ub = searchsortedlast(vals, min(ν_bound.upper, ν_div_T_bound.upper * T))
```

test/absorption_comparison_funcs.jl
mabruzzo: Not exactly. For continuum absorption of He and H_2^+, Gray still defines  the defines the "continuous absorption coefficient per neutral hydrogen atom", κ = α_ν/nH_I. 

Plus, the idea of an absorption coefficient for free-free absorption is something of a murky subject. In detail, the relevant physical quantity for something like He I free-free absorption is technically a "cross-section per free-free electron" for He II and has a dimensionality of `length^5`. As we discussed earlier today, people sometimes employ the Saha equation to transform that quantity to effectively be a cross-section per He I particle.

PR #88, opened 2021-11-05
use our own constants to compute the electron scattering cross section

### LINKED ISSUE #65: Explicitly compute coefficient for electron scattering
We currently use the numerical value from Kurucz, multiplied by 10, but I think it would make more sense to directly calculate the coefficient using the quantities in `constants.jl`.

From Gray page 160:
image

I checked that this gives the same numerical value (to 10^-2) as Kurucz.  Weirdly, Gray seems to be less accurate. (Unless one of our constants is off.)

ajwheeler:
closes #65 

```julia
julia> Korg.ContinuumAbsorption.electron_scattering(1)
6.652447401049883e-25
```
✅

### DIFF
--- a/src/continuum_absorption/scattering.jl
+++ b/src/continuum_absorption/scattering.jl
@@ -46,5 +46,5 @@ wavelength dependence. It assumes isotropic scattering.  (See, e.g. Gray p 160.)
 - `nₑ::F`: number density of free electrons (in cgs)
 """
 function electron_scattering(nₑ::F) where {F<:Real} 
-    nₑ  *  8π/3 * (electron_charge_cgs^2/(electron_mass_cgs * c_cgs^2))^2
+    8π/3 * (electron_charge_cgs^2/(electron_mass_cgs * c_cgs^2))^2 * nₑ
 end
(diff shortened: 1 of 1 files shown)


src/continuum_absorption/scattering.jl
```diff
@@ -40,14 +40,11 @@ end
...
 - `nₑ::F`: number density of free electrons (in cgs)
-
-# Notes
-I adopted the formula described in section 5.12 of Kurucz (1970) and the equation in the electron
-scattering subsection of Gray (2005); the actual coefficient value comes from the latter. It turns
-out that the coefficient in Kurucz (1970) has a typo (it's a factor of 10 too large).
 """
-electron_scattering(nₑ::F) where {F<:Real} = 0.6648e-24*nₑ
+function electron_scattering(nₑ::F) where {F<:Real} 
+    nₑ  *  8π/3 * (electron_charge_cgs^2/(electron_mass_cgs * c_cgs^2))^2
```
mabruzzo, line 49:
```suggestion
    8π/3 * (electron_charge_cgs^2/(electron_mass_cgs * c_cgs^2))^2 * nₑ
```
Micro-optimization: The order of multiplication/division operations is left to right and floating point operations are not associative. Therefore Julia can't precompute the entire floating point constant as you have currently written the code (so there will be division each time the function is called). My reordering let's julia precompute the constant (alternatively, you could enclose the entire constant in parentheses)

I have verified this locally with `code_llvm`

mabruzzo (approved):
I'd make the single recommended change and then merge

Issue #83, opened 2021-10-22
continuum absorption accuracy 


ajwheeler:
![image](https://user-images.githubusercontent.com/711963/138517668-901e5963-a87a-4758-9b0a-09a07f6c8bdd.png)

There's lots to talk about in the figure comparing the continuum from korg, moog, and TS, but I think the jump in the Korg continuum at 3846\AA (the H2+ bf+ff cutoff, marked with a grey dashed line) is indicating that this source of absorption is overestimated.

Thoughts?

ajwheeler:
I'm continuing to look into this.  I'll definitely look at the MOOG/TS version, but I think a good way to go would be to update our prescription from [Bates 1952](http://articles.adsabs.harvard.edu/pdf/1952MNRAS.112...40B) to [Stancil 1994](http://articles.adsabs.harvard.edu/pdf/1994ApJ...430..360S). This will also give us coverage down to 500 Å, although the temp lower bound will go up (perhaps, we could interpolate?).  It's also 40 years more recent, and presumably much more accurate.

Also, I didn't realize that H2+ bf refers to H2 II + gamma -> H I + H II !  That should definitely be documented! (Sorry if I missed it.)

ajwheeler:
To be clear, is TS using the data form the Mihalas paper?  I can't actually find the part of the code where they do this...

ajwheeler:
OK, so the new (Stancil) implementation reproduces the continuum offset with other codes, as expected, since it matches the Bates/Gray values.  The good news is that we don't have the problematic discontinuity around 3850 \AA, where the old values stopped.
![image](https://user-images.githubusercontent.com/711963/142253994-6c72d4ba-7b7d-42f9-9605-41cdee2f41db.png)
![image](https://user-images.githubusercontent.com/711963/142254019-a8bb6447-2abc-4e5c-bbea-0e29139bd247.png)

Manually turning it off gets a closer match, but I'm now pretty convinced we aren't actually doing anything wrong RE H2plus.  It may be that something else is overestimated, or that the other codes are underestimating somthing.

ajwheeler:
I've checked that the ff contribution from H2plus is very small and does not account for the discrepancy.

ajwheeler:
[...]

PR #93, opened 2021-11-16
Updated h2plus continuum absorption

ajwheeler:
Updates the prescription for H2plus bf and ff absorption to Stancil 1994.  See #83 for context.


src/continuum_absorption/absorption_H.jl
```diff
@@ -286,7 +286,6 @@ Hminus_ff = bounds_checked_absorption(_Hminus_ff,
 
 function _H2plus_bf_and_ff(ν::Real, T::Real, nH_I_div_partition::Real, nH_II::Real)
```
mabruzzo, line 287: So I don't think you deleted the old implementation of `_H2plus_bf_and_ff`. I think you should probably delete the old definition (especially since the code used in the unit tests is a direct transcription of the data that this polynomial fits).

Honestly, I'm surprised that Julia hasn't complained about having multiple definitions for `_H2plus_bf_and_ff` with the same function signature.?

src/continuum_absorption/absorption_H.jl
```diff
@@ -313,63 +312,52 @@ function _H2plus_bf_and_ff(ν::Real, T::Real, nH_I_div_partition::Real, nH_II::R
...
-probably correct "to well within one part in ten even at the lower temperatures and [lower
-wavelengths]."
+This computes the H₂⁺ number density from those of H I and H II, since ionized molecules 
+are not included in the molecular equlibrium calculation.  Stancil provides approximate equilibrium 
+constants, K, for the molecule, but the Barklem and Collet values used elsewhere by Korg may be more 
+reliable.  Once ionized molecules are fully supported, those values should be used instead. While 
+cross sections are tabulated down to only 3150 K, the cross sections could be linearly interpolated 
+1000 K or so lower, if reliable K values are availble.
+
+Becauce n(H₂⁺) is computed on the fly, the "cross-sections" used in this function have units of
```
mabruzzo, line 350:
```suggestion
Because n(H₂⁺) is computed on the fly, the "cross-sections" used (internally) by this function have units of 
```
2 minor suggestions:
- fix the typo: "Becauce" -> "Because"
- i made a slight tweak to try to make it clear that the cross-sections are only used internally (to make it clear that it's not something most users should care about). But definitely feel free to ignore this if you disagree

mabruzzo (approved):
Looks good to me. I think you should probably address my comment about the old code. 

The rest of my comments are all nitpicks, so feel free to ignore them.

PR #94, opened 2021-11-18
isotopic scaling 

ajwheeler:


src/isotopic_abundances.jl
```diff
@@ -0,0 +1,290 @@
+# these are scraped from NIST
+# maps (atomic number, atomic weight) to abundance (abundances for an element sum to 1)
+isotopic_abundances = Dict((71, 175) => 0.97401,
```
mabruzzo, line 3: 2 minor suggestions:
1. It might make sense to introduce a simple test to ensure that for every element with isotopic abundances, that all the abundances sum to 1 (to within floating point errors).
2. Would it make sense to organize the initialization of this dictionary so that all entries for a given atomic number are grouped together? (I realize this is mostly aesthetic, but it could make it easier for people to lookup which values are used by default - definitely feel free to ignore this suggestion).

mabruzzo (approved):
Looks good to me. I just had 2 minor suggestions.

PR #97, opened 2021-12-30
new integration scheme for tau

ajwheeler:
This does several things (sorry).
- It migrates from kurucz to marcs format model atmospheres, thus avoiding the delicate "RHOX" column.
- It introduces a new scheme for calculating tau, which is more numerically stable and which rescales to match the model atmosphere at the reference wavelength.
- It keeps track of the photosphere radius (where tau_ros == 1), and scales the final flux according to that. (This is for parity with Turbospectrum.  It doesn't matter much in real-world use.)


src/transfer.jl
```diff
@@ -1,25 +1,100 @@
 """
-    _ray_approximate_transfer_integral(τ, m, b)
+    radiative_transfer(atm::ModelAtmosphere, α, S, α_ref, mu_grid)
 
-The exact solution to \$\\int (m\\tau + b) \\exp(-\\tau)\$ d\\tau\$.
+Returns the astrophysical flux at each wavelength.
+
+inputs:
+- `atm`: the model atmosphere.
+- `α`: a matric (atmospheric layers × wavelengths) containing the absoprtion coefficient
```
mabruzzo, line 8:
```suggestion
- `α`: a matrix (atmospheric layers × wavelengths) containing the absoprtion coefficient
```

src/synthesize.jl
```diff
@@ -92,59 +92,49 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
...
-        α_cntm_vals = total_continuum_absorption(sorted_cntmνs, layer.temp,
-                                                 layer.electron_number_density,
-                                                 ns, partition_funcs)
-        reverse!(α_cntm_vals)
-        LinearInterpolation(cntmλs, α_cntm_vals) # LinearInterpolation needs ranges to be increasing
-    end
+    sorted_cntmνs = c_cgs ./ reverse(cntmλs) #frequencies at which to calculate the continuum
 
+    α_cntm = Vector(undef, length(atm.layers))     #vector of continuum-absorption interpolators
+    α5 = Vector{α_type}(undef, length(atm.layers)) #reference absorption at each layer
```
mabruzzo, line 104:
```suggestion
    α5 = Vector{α_type}(undef, length(atm.layers)) #each layer's absorption at reference λ (5000 Å)
```
I know this seems pedantic, but I think this tweak should avoid confusion (I also confirmed that the line is still under 100 characters)

ajwheeler:
Just confirmed that this is a serious performance regression for spherical atmospheres, dominating the runtime for the apogee linelist.  I'm going to try to speed it up, so don't merge.

image

ajwheeler:
OK, this could probably be faster, but it's now a slight improvement on the performance of the main branch rather than a regression.
image

src/transfer.jl
```diff
@@ -1,120 +1,210 @@
...
+Compute exactly the solution to the transfer integral obtained be linearly interpolating the source 
+function, `S` across optical depths `τ`, without approximating the factor of exp(-τ)
+"""
+function ray_transfer_integral(τ, S)
+    I = 0
+    for i in 1:length(τ)-1
+        @inbounds m = (S[i+1] - S[i])/(τ[i+1] - τ[i])
+        @inbounds b = S[i] - m*τ[i]
+        @inbounds I += (_ray_approximate_transfer_integral(τ[i+1], m, b) - 
+                        _ray_approximate_transfer_integral(τ[i], m, b))
```
mabruzzo, line 144: Since `ray_transfer_integral` is the most expensive part of spherical transfer, it's probably worth exploring whether `_ray_approximate_transfer_integral` is being correctly inlined at some point. 

I'm a little worried that the compiler might not realize that it can rewrite this for loop as follows (and evaluating twice as many exponential functions as it needs to):

```julia
    next_exp_negτ = exp(-τ[1])
    for i in 1:length(τ)-1
        @inbounds m = (S[i+1] - S[i])/(τ[i+1] - τ[i])
        @inbounds b = S[i] - m*τ[i]
        cur_exp_negτ = next_exp_negτ
        @inbounds next_exp_negτ = exp(-τ[i+1])
        @inbounds I += (-next_exp_negτ * (m*τ[i+1] + b + m) +  
                        cur_exp_negτ * (m*τ[i] + b + m))
    end
```

If you have a spare moment, would you mind copying the preceding snippet into the code and seeing if the code gets any faster? If it does speed up the code, we might be able to reproduce the same effect with `@inline` (although I remember that the inline macro was temperamental, so it would be better to test things by directly overwriting the loop)

mabruzzo: Actually, if we wanted to take this to the logical extreme, one could imagine entirely replacing `ray_transfer_integral` with the following (but I'm not sure it's worth doing given that it's not obvious where this comes from).

```julia
function ray_transfer_integral(τ, S)
    I = 0
    next_exp_negτ = exp(-τ[1])
    for i in 1:length(τ)-1
        @inbounds m = (S[i+1] - S[i])/(τ[i+1] - τ[i])
        cur_exp_negτ = next_exp_negτ
        @inbounds next_exp_negτ = exp(-τ[i+1])
        @inbounds I += (-next_exp_negτ * (S[i+1] + m) +  
                        cur_exp_negτ * (S[i] + m))
    end
    I
end
```

src/transfer.jl
```diff
@@ -1,120 +1,210 @@
...
+        lowest_layer_indices[μ_ind] = i
+
+        integrand_factor[μ_ind, 1:i] = @. radii[1:i] ./ sqrt(radii[1:i]^2 - b^2) * τ_ref_α_ref[1:i]
+    end
+
+    #preallocations
+    I = Matrix{el_type}(undef, size(α, 2), length(μ_surface_grid)) #the surface intensity
+    integrand = Vector{el_type}(undef, size(α, 1))                 #integrand of τ integral 
+    τ_λ = Vector{el_type}(undef, size(α, 1))                       #optical depth at a particular λ
+    for λ_ind in 1:size(α, 2), μ_ind in 1:length(μ_surface_grid)
```
mabruzzo, line 89: Are we iterating over λ in the outer loop? 

I'd probably add a comment clarifying this since this fairly foreign notation

src/transfer.jl
```diff
@@ -1,120 +1,210 @@
...
-            #if the ray never leaves the model atmosphere, include 
-            #the contribution from the other side of the star
-            τ_prime = τ[end:end, :] .+ [cumsum(reverse(dτ[2:end, :], dims=1), dims=1) 
-                                        ; zeros(size(τ[end:end, :]))]
-            for j in 1:size(τ, 2) #Iterate over wavelength
-                I[j, μ_ind] += transfer_integral(τ_prime[:, j], S[1:i, j]; plane_parallel=false)
-            end
-        else #otherwise assume I=S at atmosphere lower boundary
-            I[:, μ_ind] += @. exp(-τ[end, :]) * S[end, :]
+        cumulative_trapezoid_rule!(τ_λ, log_τ_ref, integrand, i)
```
mabruzzo, line 94: potential micro-optimization: `log_τ_ref` is identical every time you compute this, right? If so, you could precompute delta_log_τ_ref ahead of time and just pass that into the loop instead.

(Feel free to ignore this)

src/transfer.jl
```diff
@@ -1,120 +1,210 @@
...
+    end
+
+    #preallocations
+    I = Matrix{el_type}(undef, size(α, 2), length(μ_surface_grid)) #the surface intensity
+    integrand = Vector{el_type}(undef, size(α, 1))                 #integrand of τ integral 
+    τ_λ = Vector{el_type}(undef, size(α, 1))                       #optical depth at a particular λ
+    for λ_ind in 1:size(α, 2), μ_ind in 1:length(μ_surface_grid)
+        i = lowest_layer_indices[μ_ind]
+        for k in 1:i #I can't figure out how to write this as a fast one-liner
+            integrand[k] = α[k, λ_ind] * integrand_factor[μ_ind, k]
```
mabruzzo, line 92: micro-optimization: you should swap the axes order of `integrand_factor[μ_ind, k]` so that atmospheric layer indexes the fast-access axis (again, feel free to ignore this suggestion)

src/transfer.jl
mabruzzo, line 94: Maybe consider adding a comment like the following?
```julia
        cumulative_trapezoid_rule!(τ_λ, log_τ_ref, integrand, i) # compute τ_λ
```

I totally get it if you choose not to do so (since it is julia's convention for functions ending in `!` to mutate the first argument)

src/transfer.jl
```diff
@@ -1,120 +1,210 @@
...
+        if i < length(radii)
+            #if the ray never leaves the model atmosphere, include the contribution from the 
+            #other side of the star.  Calculate the optical depths you get by reversing the τ_λ.
+            #This is less accurate than actually integrating to find τ, but the effect is small.
+            #This should probably by audited for off-by-one errors.  It's also inneficient.
+            τ_prime = τ_λ[i] .+ [cumsum(reverse(diff(view(τ_λ,1:i)))) ; 0]
+            I[λ_ind, μ_ind] += ray_transfer_integral(view(τ_prime, 1:i), view(S,1:i,λ_ind))
+        else #otherwise assume I=S at atmosphere lower boundary.  This is a _tiny_ effect.
+            I[λ_ind, μ_ind] += exp(-τ_λ[end]) * S[end, λ_ind]
         end
```
mabruzzo, line 106: It doesn't seem like branch-misses are a problem (then again, it's never been clear to me how they would manifest in the flamegraph), but I'm pretty sure that this branch does not need to exist within the innermost loop.

If it ever became necessary, I think you could subdivide `for μ_ind in 1:length(μ_surface_grid)` into 2 separate branches that execute the different branches.

ajwheeler: This is an excellent point.

src/transfer.jl
ajwheeler: I'm going to ignore this one for now as I think it's a sub-1% effect and I don't want to encumber other optimizations.

src/transfer.jl
ajwheeler: I think this is a good idea.  It also concords nicely with my (arguably wrong) preference for few, more complicated, functions.

src/transfer.jl
```diff
@@ -1,120 +1,211 @@
...
-        
-        for j in 1:size(τ, 2) #iterate over wavelength
-            I[j, μ_ind] = transfer_integral(τ[:, j], S[1:i, j]; plane_parallel=false)
+        lowest_layer_indices[μ_ind] = i
+
+        integrand_factor[1:i, μ_ind] = @. radii[1:i] ./ sqrt(radii[1:i]^2 - b^2) * τ_ref_α_ref[1:i]
+    end
+
+    #preallocations
+    I = Matrix{el_type}(undef, size(α, 2), length(μ_surface_grid)) #the surface intensity
```
mabruzzo, line 86: Sorry, I missed this before. You should also consider flipping the axes order on this matrix so that `μ_surface_grid` is along the fast-access-axis. This should make access in the following loop. It will also make the calculation of astrophysical flux faster.

Just a heads up, if you do this, you're going to need to modify line 111 to read `F = 2π * [Korg.trapezoid_rule(μ_surface_grid, μ_surface_grid .* I) for I in eachcol(I)]`

mabruzzo (approved):
Looks great to me. Sorry for not grouping my comments together. in a single review

src/transfer.jl
ajwheeler: Wait, don't you also need the contribution from `b` at the first and last indices?

mabruzzo: Unless I did something incorrectly, I don't think so. I literally just substituted the equation for b in and things cancelled out, a lot

src/transfer.jl
ajwheeler: I've noted it in a comment.

src/transfer.jl
ajwheeler: Ah, I see. Fixed.

ajwheeler: This is about a 20% speedup, btw.  (Sorry for the message spam.)

mabruzzo: Okay so here's the sequence of algebraic steps I followed. I started with this function:

```julia
function ray_transfer_integral(τ, S)
    I = 0
    next_exp_negτ = exp(-τ[1])
    for i in 1:length(τ)-1
        @inbounds m = (S[i+1] - S[i])/(τ[i+1] - τ[i])
        @inbounds b = S[i] - m*τ[i]
        cur_exp_negτ = next_exp_negτ
        @inbounds next_exp_negτ = exp(-τ[i+1])
        @inbounds I += (-next_exp_negτ * (m*τ[i+1] + b + m) +  
                        cur_exp_negτ * (m*τ[i] + b + m))
    end
    I
```

1. I literally just extracted the following line:
```julia
    I += (-next_exp_negτ * (m*τ[i+1] + b + m) +  
          cur_exp_negτ * (m*τ[i] + b + m))
```
2. I substituted in `b = S[i] - m*τ[i]`
```julia
    I += (-next_exp_negτ * (m*τ[i+1] + (S[i] - m*τ[i]) + m) +  
          cur_exp_negτ * (m*τ[i] + (S[i] - m*τ[i]) + m))
```
3. I reorganized the statement slightly:
```julia
    I += (-next_exp_negτ * (m*(τ[i+1]-τ[i]) + S[i] + m) +  
          cur_exp_negτ * (S[i] + m))
```
4. I replaced `m*(τ[i+1]-τ[i])` with `(S[i+1] - S[i])`:
```julia
    I += (-next_exp_negτ * ((S[i+1] - S[i]) + S[i] + m) +  
          cur_exp_negτ * (S[i] + m))
```
5. After a final cancellation, we're left with:
```julia
    I += (-next_exp_negτ * (S[i+1] + m) +  
          cur_exp_negτ * (S[i] + m))
```

Using this result, you can rewrite the function I started with as:
```julia
function ray_transfer_integral(τ, S)
    I = 0
    next_exp_negτ = exp(-τ[1])
    for i in 1:length(τ)-1
        @inbounds m = (S[i+1] - S[i])/(τ[i+1] - τ[i])
        cur_exp_negτ = next_exp_negτ
        @inbounds next_exp_negτ = exp(-τ[i+1])
        @inbounds I += (-next_exp_negτ * (S[i+1] + m) +  
                        cur_exp_negτ * (S[i] + m))
    end
    I
end
```

ajwheeler:
Thanks for the suggestions/catching my mistakes.  They took the runtime for the apogee linelist for a giant star from ~1.7s to ~1.4s.

mabruzzo:
Wow, that's honestly a much bigger performance bump than I was expecting. Would you mind posting the updated flamegraph? I'm honestly very curious to see what it looks like now

ajwheeler:
image

PR #99, opened 2022-01-24
Metal ff absorption for positively charged species

mabruzzo:
**Overview**
I finally got around to digging through the partial progress I had made back in October for the general free-free metal opacity function, for interactions involving positively ionized metal ions and I'm posting it here. 

When I was first looking into this, it took me a while to understand what people were doing, but here it is:
- According to Table I of [Gustafsson+ (2008)](https://ui.adsabs.harvard.edu/abs/2008A%26A...486..951G/abstract), MARCS (and turbospectrum, by extension) use they hydrogenic approximation for most species (the value of `Z` depends on the species).
- For a subset of species, they employ correction terms from [Peach (1970)](https://ui.adsabs.harvard.edu/abs/1970MmRAS..73....1P/abstract). The species subset seems to be motivated by which ones have Departure terms under normal atmospheric conditions and which ones are most abundant for the sun.
- Part of my confusion arose from the fact that most of the literature on free-free absorption talks don't talk about the interaction in terms of the species that participate in the interaction (e.g. Si I free-free absorption involves a free-electron and a Si II ion). 
    - Furthermore, most of the equations are actually written in terms of the species that the interaction is named after, and implicitly include terms from the Saha equation to convert that to the number density of the involved species (e.g. They would write the equations for Si I free-free absorption using the number density in terms of the number density of Si I instead of Si II. The equations would also involve terms from the Saha equation to implicitly convert to the number density of Si II).
    - In hindsight, it may have been better to just embrace this convention.

**Departure terms:**
- Peach (1970) provides tables of `D(T,σ)` for various species, where σ is the energy of the absorbed photon in units of Rydbergs (it's probably worth a sanity check to see whether this is Rydbergs for the selected species, or the general Rydberg unit).
- To use the correction term, we just multiply `(1+D(T,σ))` by the hydrogenic free-free absorption coefficient.
- At this point, it's a little unclear to me how exactly we handle C II free-free absorption (the interaction involving C III), since the table splits it up by "parent term." (I had actually understood all of this back in October, but never wrote it down).
- It's not obvious to me how we should interpolate `D(T,σ)` outside of the table's bounds
- Peach (1970) provides additional tables that can be used to check that we are doing things correctly. I did this already in `test/ff.jl` for He I free-free absorption. There isn't a great reason that this was separate from the other tests (other than to make things easy as I started on this).

**Overall**

I threw together the `all_metal_ff_absorption` function to try to make the general procedure for computing the free-free continuum absorption as explicit as possible. (However, I think we may need to divide the number density of each species by the partition functions).

But, I think that the best way forward to finishing things up for a paper might be:
- to ignore the departure terms altogether and just use the Hydrogenic approximation for all species. Then we could go back later and finish up with the departure terms later if we think they're necessary. To be honest, my intuition tells me that the inclusion of departure term is negligible ( < 1% effect to the total continuum). As an aside, I don't think that the Cloudy uses any departure terms at all. 
- To achieve that easily, we could just make `departure_dict` an empty dictionary within the `all_metal_ff_absorption` function.

From an organizational standpoint:
- I also think that we probably delete the separate function used for `He_II_ff` and just handle that term with the rest of the metals. But, it probably still makes sense to handle `H_I_ff` separately.
- It might make sense to move all routines for computing free-free absorption for positively charged ions to a single file (including the routine for  ``hydrogenic_ff_absorption``, and ``H_I_ff`` to the same file as these new routines.

**Cleaning up the code**

This was all very-much a work-in-progress and consequently the code is a little messy. I'm not actually sure that this code will currently execute.

Having just pulled this all together, I'm feeling motivated to help you wrap this up. If I get more time this week I would be tempted to come back and clean things up more. But I definitely don't want to get in your way. Just let me know if you want me to keep my hands off the code. (Either way, I'm happy to clarify any questions you might have).

Let me know your thoughts on all of this


src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+Saha Equation relates ``\\alpha_{\rm ff}``'s dependence on ``n_e`` and ``n_i`` to ``n_{i-1}`` and 
+``T``.  Gray (2005) follows a similar convention when describing free-free absorption.
+
+As an aside, a ``D(T, \\sigma)`` that is zero everywhere, is equivalent to just using a hydrogenic
+approximation.
+"""
+function metal_ff_absorption_departure(ν::Real, T::Real, Z::Integer, ni::Real, ne::Real, departure)
+    σ = ν/ Z^2 * (hplanck_eV / Rydberg_eV)
+    hydrogenic_ff_absorption(ν, T, Z, ni, ne) * (1 + departure(T, σ))
+end
```
ajwheeler, line 67: I think refactoring this into something which returns `1 + departure(T, sigma)` might be preferable, in that it would make things more explicit higher in the stack (it's also theoretically more flexible, but I don't think that's a big deal).

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+
+
+# should we be loading this from a hdf5 table?
```
ajwheeler, line 70: imho it's preferable to do this for the sake of package load time when tables are small.  I do think it would be a good idea to put these in a separate ".jl" though.

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+const _departure_term_He_I_ff = begin
+    # this comes from table III of Peach 1970 for neutral Helium
```
ajwheeler, line 72: I realized we can use `let` blocks (with no bindings) instead of `begin`s to actually introduce local scope.

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
+using Interpolations: LinearInterpolation, Throw
+
+# We have implemented the free-free metal departure terms (from Peach 1970) for the same species as
+# MARCS/Turbospectrum (see Table 1 of Gustafsson+ 2008).
```
ajwheeler, line 4: We should enumerate these somewhere.  It's
 - He I
 - C I
 - C II
 - Mg I
 - Si I
 
Right?

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+                  -0.048 -0.042 -0.020  0.006  0.038  0.123;
+                  -0.045 -0.040 -0.019  0.006  0.039  0.124;
+                  -0.039 -0.035 -0.016  0.008  0.042  0.128;
+                  -0.032 -0.030 -0.012  0.011  0.046  0.133]
+
+    LinearInterpolation((T_vals, σ_vals), table_vals, extrapolation_bc=Throw())
+end
+
+_Mg_I_ff(ν::Real, T::Real, ndens_MgII::Real, nₑ::Real) =
+    metal_ff_absorption_departure(ν, T, 1, ndens_MgII, nₑ, _departure_term_Mg_I_ff)
```
ajwheeler, line 360: I'm a bit skeptical of the utility of these functions.  Is it just about wrapping globals wherever possible?

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+
+
+"""
+    metal_ff_absorption(ν, T, Z, ni, ne)
+
+Computes the free-free linear absorption coefficient (in cm⁻¹) of a metal species for which a 
+departure term is specified.
```
ajwheeler, line 367:
```suggestion
Computes the free-free linear absorption coefficient (in cm⁻¹) of a metal species for which a 
departure term is not specified.
```

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+This function simply wraps the hydrogenic free-free calculation.
+
+# Arguments
+- `ν`: frequency in Hz
+- `T`: temperature in K
+- `Z::Integer`: the net charge of the ion species (that participates in the interaction).
+- `ni`: the number density of the ion species (that participates in the interaction) in cm⁻³.
+- `ne`: the number density of free electrons.
+"""
+metal_ff_absorption(ν, T, Z, ni, ne) = hydrogenic_ff_absorption(ν, T, Z, ni, ne)
```
ajwheeler, line 385: I'm kinda skeptical about the utility of this as well.  I think it's preferable to make things explicit higher in the call stack and avoid defining extra functions.

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+
+
+
+"""
+    all_metal_ff_absorption(ν::Real, T::Real, number_densities::Dict, ne::Real,
+                            exclude_species_with_departure_terms::Bool = false)
+
+Computes the free-free linear absorption coefficient (in cm⁻¹) of metal species. This only includes
+free-free absorption for species with a net positive charge of 1, 2, or 3. It also skips Hydrogenic
+atoms.
```
ajwheeler, line 395: I don't think He is considered a metal.  I'm ambivalent about whether it should be included here, but if it is it should probably be mentioned explicitly.

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+                                 exclude_species_with_departure_terms::Bool = false)
+
+    error("I think we may need the user to pass in the partition functions for every species")
+
+    # recall, name of the free-free interaction differs from the name of the participating species
+    departure_dict = Dict([(species"He_II", _He_I_ff),
+                           (species"C_II",  _C_I_ff),
+                           (species"C_III", _C_II_ff),
+                           (species"Si_II", _Si_I_ff),
+                           (species"Mg_II", _Mg_I_ff)])
```
ajwheeler, line 422: One possible way to refactor this is to define this dictionary in the file where the `D` values are tabulated.  Alternatively, you could do a function `departure_from_hydrogenic(spec::Species)` with `if else`s to switch between the species.  You could then hide the interpolator objects behind a module and export that function only.  In this vision, this would replace `metal_ff_absorption_departure` and all the info in that docstring would be in the docstring for this function as well.

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+
+    ndens_Z1 = 0.0
+    ndens_Z2 = 0.0
+    ndens_Z3 = 0.0
+
+    α_out = 0.0 * ν
+
+    for (k,ndens) in number_densities
+        if k in [species"H_II", species"He_III", species"Li_IV"]
```
ajwheeler, line 431: Not sure if this line will be in the final code, but if it is we should note that Li IV is future-proofing, not something that will actually come up in the current version of Korg.

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+!!! note
+    This mostly exists for demonstrative purposes.
+
+    It might be useful to adopt this function for long-term use for efficiency reasons (even though
+    it departs from the format of all other continuum absorption calculations).  
+
+"""
+function all_metal_ff_absorption(ν::Real, T::Real, number_densities::Dict, ne::Real,
+                                 exclude_species_with_departure_terms::Bool = false)
+
```
ajwheeler, line 414: Why would we want this option?

src/continuum_absorption/absorption_ff_metal.jl
ajwheeler, line 422: This isn't important at all and I'm not requesting a change, but you might want to know that `species"C II"` works as well ( ` ` not `_`)

src/continuum_absorption/absorption_ff_metal.jl
```diff
@@ -0,0 +1,458 @@
...
+                ndens_Z3 += ndens
+            end
+        end
+    end
+
+    α_out += metal_ff_absorption(ν, T, 1, ndens_Z1, ne)
+    α_out += metal_ff_absorption(ν, T, 2, ndens_Z2, ne)
+    α_out += metal_ff_absorption(ν, T, 1, ndens_Z3, ne)
+    α_out
+end
```
ajwheeler, line 457: I don't understand how this function works...  I will bother you about it in person.

ajwheeler (commented):
You might want to read these all before you start replying, sorry I didn't organize my thoughts better.

> At this point, it's a little unclear to me how exactly we handle C II free-free absorption (the interaction involving C III), since the table splits it up by "parent term." (I had actually understood all of this back in October, but never wrote it down).

Is this the electron state of the C II?  Weird.

> It's not obvious to me how we should interpolate D(T,σ) outside of the table's bounds

I vote for `Flat()`.

> to ignore the departure terms altogether and just use the Hydrogenic approximation for all species. Then we could go back later and finish up with the departure terms later if we think they're necessary. To be honest, my intuition tells me that the inclusion of departure term is negligible ( < 1% effect to the total continuum). As an aside, I don't think that the Cloudy uses any departure terms at all.

How much is left to do?  It seems like you've already done almost all the work?  Is it just the business with C II or is there something else?

src/continuum_absorption/absorption_ff_metal.jl
mabruzzo: I totally agree about this. I was just trying to get what I had out of my hands and be as explicit as possible about everything in light of the fact that not everything is polished.

ajwheeler:
@mabruzzo this is now ready for review.

ajwheeler:
I've tweaked your example function into one to handle all ff absorption except that by the negative ions.  I imagine you don't like the idea of handling hydrogen and helium here too.  I'm happy to discuss if you think it's a mistake.

I've set the interpolators to return 0 (no correction) outside of their support, because some of them turn on at quite high temperatures (10,000 K), and we don't want to apply those departure coefficients at lower temps

>  Peach (1970) provides tables of `D(T,σ)` for various species, where σ is the energy of the absorbed photon in units of Rydbergs (it's probably worth a sanity check to see whether this is Rydbergs for the selected species, or the general Rydberg unit).

It's the general Rydberg unit.  [See equation (5) of Peach 1967](https://articles.adsabs.harvard.edu/pdf/1967MmRAS..71....1P). 

>At this point, it's a little unclear to me how exactly we handle C II free-free absorption (the interaction involving C III), since the table splits it up by "parent term." (I had actually understood all of this back in October, but never wrote it down).

I did a lot of reading about this, and I think I now understand how we would do it correctly.  That said, I'm not sure that it's worth the added code complexity for a correction to what is already a tiny effect.  For now, we are not doing C II ff, although the tables are there but commented out.


I don't think we need to be adjusting for the fraction of atoms in the ground state.  Here's why:
 - I tested our C I ff values vs Peach's.  The largest difference was ~5%.
 - If we were to adjust for the fraction of C II atoms in the ground state, the calculation is 
 ```
n(C II ground)     g      2
-------------- = ----- = ---- ~= 0.3
   n(C II)        U(T)    ~6
 ```
 which is too large a difference.

src/ContinuumAbsorption/absorption_ff_positive_ion.jl
```diff
@@ -0,0 +1,74 @@
...
+   should be callables taking temperature and (photon energy / RydbergH / Zeff^2).  See 
+   [Peach1970](@ref) for details.
+
+!!! note
+    A free-free interaction is named as though the species interacting with the free electron had 
+    one more bound electron (in other words it's named as though the free-electron and ion were 
+    bound together). The `Z` argument and `ni` values should be specified for the species that 
+    actually participates in the reaction. For example:
+    - Si I ff absorption: `ni` holds the number density of Si II, and `Z=1` (net charge of Si II)
+    - Si II ff absorption: `ni` holds the number density of Si III, and `Z=2` (net charge of Si III)
```
mabruzzo, line 28:
```suggestion
!!! note
    A free-free interaction is named as though the species interacting with the free electron had 
    one more bound electron (in other words it's named as though the free-electron and ion were 
    bound together). To make matters more confusing, some sources present the free-free linear 
    absorption coefficient's formula in terms of the number density of the species in the 
    interaction's name (by including the Saha equation in the formula).

    To compute the absorption for a given free-free interaction, this function accesses elements 
    from the dictionary arguments that are associated with the species that participates in the 
    interaction. For example:
    - Si I ff absorption: uses the number density of Si II (`number_densities[species"Si II"]`) and
      checks `departure_coefficients[species"Si II"]` for departure coefficients.
    - Si II ff absorption: uses the number density of Si III (`number_densities[species"Si III"]`)
      and checks `departure_coefficients[species"Si III"]` for departure coefficients.
```
This topic is confusing and I think it may help to add further clarification like the above

src/ContinuumAbsorption/absorption_ff_positive_ion.jl
```diff
@@ -0,0 +1,74 @@
...
+# Arguments
+- `ν`: frequency in Hz
+- `T`: temperature in K
+- `number_densities` is a `Dict` mapping each `Species` to its number density.
+- `ne`: the number density of free electrons.
+- `departure_coefficients` (optional, defaults to [Peach1970](@ref)): 
+   a dictionary mapping species to the departure coefficients for the ff process it participates in 
+   (e.g. C II maps to the C III ff departure coefficients--see note below).  Departure coefficients 
+   should be callables taking temperature and (photon energy / RydbergH / Zeff^2).  See 
+   [Peach1970](@ref) for details.
```
mabruzzo, line 20:
```suggestion
   (e.g. `species"C II"` maps to the C III ff departure coefficients--see note below).  Departure
   coefficients should be callables taking temperature and (photon energy / RydbergH / Zeff^2).  
   See [Peach1970](@ref) for details.
```

mabruzzo (commented):
Looks Great to Me!

I have some minor suggestions comment/docstring suggestions (mostly to provide further explanation). After you address those, this should be good to go.

One other thought (that I mentioned in person): it may make sense to move the hydrogenic free-free function (and gaunt factor functions) into `src/ContinuumAbsorption/absorption_ff_positive_ion.jl`. But no need to do that in this PR

(also it won't let me approve the PR since I originally created it)

src/ContinuumAbsorption/absorption_ff_positive_ion.jl
```diff
@@ -0,0 +1,74 @@
...
+            else
+                error("triply+ ionized species not supported")
+            end
+        end
+    end
+
+    #add contributions from species for which we use the uncorrected hydrogenic approximation
+    @. α_out[idx] += hydrogenic_ff_absorption(νs[idx], T, 1, ndens_Z1, ne)
+    @. α_out[idx] += hydrogenic_ff_absorption(νs[idx], T, 2, ndens_Z2, ne)
+    ;
```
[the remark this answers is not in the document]
ajwheeler: It's syntactically associated with the line above, preventing it the function from returning a value.  It could be removed with no impact on the code's functioning.

src/ContinuumAbsorption/absorption_H.jl
```diff
@@ -44,29 +46,6 @@ H_I_bf = bounds_checked_absorption(_H_I_bf; ν_bound = Interval(0, Inf),
 
 _H_I_ff(ν, T, nH_II, ne) = hydrogenic_ff_absorption(ν, T, 1, nH_II, ne)
```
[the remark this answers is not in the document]
ajwheeler: Nope.  Just deleted it.

mabruzzo (commented):
Looks Great!

PR #103, opened 2022-02-28
add back test; does it work?

ajwheeler:


ajwheeler:
I think maybe we need to wait for a new patch version.

ajwheeler:
I triggered the test suite, and everything passed: https://github.com/ajwheeler/Korg.jl/actions/runs/3062644623/jobs/4943852122.

Next time, we should probably use `@test_broken` instead of `@test_skip` since it will error when the test passes.

PR #104, opened 2022-03-01
add 1D cubic spline functionality

ajwheeler:
Here, I've ripped some code from DataInterpolations, which had a nice cubic spine implementation but has super heavy dependencies.  I've also been able to simplify it a bit.

Places where `Interpolations.LinearInterpolation is used` and whether I've replaced them with this spine implementation:
- stark profiles of H lines.  **2-dim, can't replace**
- continuum absorption coefficient **o-go because solving (`\`ing) `LinearAlgebra.Tridiagonal`s doesn't support dual numbers this will probably be remedied eventually**
- interpolating the continuum for rectification: **see below**
- `_Hminus_bf_cross_section_interp` **2-dim, can't replace**
- metal bf cross sections **2-dim, can't replace**
- hydrogenic gaunt factor **2 dim, can't replace**
- stancil data **2-dim, can't replace**
- there's something in absorption comparison funcs as well but I didn't look at it.

edit: actually we will run into the same autodiff problems with rectification when doing inference.  I'll roll that back.

Summary: this is only replacing `LinearInterpolation` for partition funcs/equilibrium constants, but it solves the problems we were having there (see plots below).  A differentiable implementation and a 2D implementation would be very useful to have still. 

TODO
 - docstrings
 - tests


ajwheeler:
image
image

src/CubicSplines.jl
```diff
@@ -0,0 +1,72 @@
+module CubicSplines
+using LinearAlgebra
+
+#=
+This module is contained modified functions from DataInterpolations.jl (license below).
+We aren't depending on DataInterpolations because it's dependencies are very heavy.
```
mabruzzo, line 6:
```suggestion
This module contains modified functions from DataInterpolations.jl (license below).
We aren't depending on DataInterpolations because its dependencies are very heavy.
```

src/CubicSplines.jl
```diff
@@ -0,0 +1,72 @@
...
+In the future, hopefully we can eliminate interpolation code and depend on a widely-used
+well-tested and lightweight library.
+
+Note, I've swapped the order of t and u (t is the abscissae/x-values, u are the y-values). The other 
+major change is that I've simplified the types. We might at some point want to add analytic 
+derivatives (DataInterpolations has these), but since ForwardDiff doesn't use ChainRules, there is 
+no way to get them used in autodiff (untill/unless you use a different library.)
```
mabruzzo, line 13:
```suggestion
no way to get them used in autodiff (until/unless you use a different library.)
```

mabruzzo (approved):
Looks good to me! I've got 2 minor suggestions (for typos). Merge whenever

Out of curiosity, does this introduce any performance regressions? I only ask because natural `log` is of similar expense to `exp` (I realize this is an unavoidable problem, I'm mostly just curious).

PR #106, opened 2022-03-25
compute stark profile for Halpha Hbeta and Hgamma

ajwheeler:
Convolve with Barklem p-d approximation for resonant broadening by adding.  (this obviously inexact, but it's better than nothing)

This change gives us better agreement with turbospectrum, but more importantly, it seems to be match data better:
image
(This is HD49933)

### DIFF
--- a/src/line_absorption.jl
+++ b/src/line_absorption.jl
@@ -186,6 +186,9 @@ function hydrogen_line_absorption(λs, T, nₑ, nH_I, UH_I, hline_stark_profiles
 
         #compute profile with either self or stark broadening
         Hmass = get_mass(Formula("H"))
+
+        #if it's Halpha, Hbeta, or Hgamma, add the H broadening as a pure lorentzian. 
+        #(convolve be adding)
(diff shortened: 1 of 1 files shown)


src/line_absorption.jl
```diff
@@ -186,6 +186,9 @@ function hydrogen_line_absorption(λs, T, nₑ, nH_I, UH_I, hline_stark_profiles
 
         #compute profile with either self or stark broadening
         Hmass = get_mass(Formula("H"))
+
+        #if it's Halpha, Hbeta, or Hgamma, add the H broadening as a pure lorentzian. 
+        #(convolve be adding)
```
mabruzzo, line 191:
```suggestion
        #(convolve by adding)
```

mabruzzo (approved):
Looks good to me.

To clarify: for Halpha, Hbeta and Hgamma, you are adding the profile involving ABO parameters and the stark broadening parameters together. For transitions from n>5 to n = 2, you are just using the stark broadening parameters?

If so, you might want to be a little more explicit that this is what is happening in the comment

PR #109, opened 2022-05-07
swap out Hminus bf cross section data

ajwheeler:
Per reviewer's request (which I agree with).

I've also tweaked the way the Stark profiles are loaded, and kept things closer to where they are used.  It requires much less jumping around for someone reading the code, and it's consistent with how I'm loading and using an HDF5 file here.

image
image


src/continuum_absorption/absorption_H.jl
```diff
@@ -1,5 +1,5 @@
 using Interpolations: LinearInterpolation, Throw
-using StaticArrays: SA
+using HDF5: h5read
 
 using ..ContinuumAbsorption: hydrogenic_bf_absorption, hydrogenic_ff_absorption, ionization_energies
```
mabruzzo, line 5: Can you update line 7 to read `const _H⁻_ion_energy = 0.754204 # eV (value adopted in McLaughlin)`

test/continuum_absorption.jl
```diff
@@ -130,89 +131,24 @@ function _Hminus_bf_cross_section_Gray(λ::Real)
...
-end
-
 
-@testset "H⁻ bound-free absorption" begin
-    @test check_Hminus_bf_values(0.0025)
     @testset "Gray (2005) Fig 8.5$panel comparison" for panel in ["a", "b", "c"]
         calculated, ref = Gray_opac_compare.Gray05_comparison_vals(panel,"Hminus_bf")
         @test all(calculated .≥ 0.0)
-        @test all(abs.(calculated - ref) .≤ Gray_opac_compare.Gray05_atols[panel])
-    end
```
mabruzzo, line 174: It might be nice to add some version of this test back in, just to check that we get a sensible non-zero value between the lowest tabulated frequency from McLaughlin+ 2017 and the ionization energy. But, it's definitely not necessary if you don't want to do it (at the very least, we should probably check that it works manually)

test/continuum_absorption.jl
```diff
@@ -103,20 +103,21 @@ end
     @testset "Gray (2005) Fig 8.5$panel comparison" for panel in ["b", "c"]
         calculated, ref = Gray_opac_compare.Gray05_comparison_vals(panel,"Hminus_ff")
         @test all(calculated .≥ 0.0)
-        @test all(abs.(calculated - ref) .≤ Gray_opac_compare.Gray05_atols[panel])
+        @test assert_allclose(calculated, ref; rtol=0, atol=0.03)
     end
 end
 
 """
     _Hminus_bf_cross_section_Gray(λ)
```
mabruzzo, line 111: Since the Wishart data is already digitized, it might make slightly more sense to use it here than the Gray polynomials. But, I don't want to make more work for you (and it really doesn't matter at the end of the day)

mabruzzo (approved):
So you can ignore what I said the other day about this change is probably not an important effect. I was confused and thought this PR was switching the H⁻ free-free opacity source from using the Grey polynomial. I'm sure you're right, this change is probably important.

While looking this over, I noticed that you had a non-zero minimum frequency bound for this new function. I was reminded that I encountered a similar scenario when I first implemented the Wishart table. In both cases the minimum values in the table is very close to the ionization energy (apparently it's called the electron affinity for negative ions) and below that energy, the cross-section should be zero (by definition). 

This really isn't important, but it would be nice to say that there is effectively no wavelength bounds on this absorption source. So, I dug around in McLaughlin+ 2017. Buried in the text, I found a comment about how they used `σ = 460.8*(Eᵧ - E₀)^1.5 Mb` for `Eᵧ < 0.7678 eV` (`E₀` is the ionization energy). I loaded the entries from the table that had these energies and was able to confirm that entries are extremely consistent with this relation.

Therefore, I have suggested a few changes for how this could be implemented (again this effect is very unimportant, it's just nice to have for completeness).

If you don't feel like implementing those suggested change, I'm happy to open a separate PR.

test/continuum_absorption.jl
mabruzzo: To clarify, I was talking about the extreme wavelength test.

test/continuum_absorption.jl
ajwheeler: Yeah, this occurred to me but I decided it wasn't worth the effort.

test/continuum_absorption.jl
ajwheeler: This is a good idea.  I've written a new test.

PR #111, opened 2022-05-15
Add support for neutral free-free absorption involving some metals/molecules

mabruzzo:
I was feeling motivated yesterday, so I decided to introduce support for some free-free absorption coefficients involving neutral metals and ions. Most/All of these are probably unimportant. They are mostly included just because Turbospectrum has them.

In detail:
- I added support for Ominus, COminus, and H2minus.
- Most of the code is in place to support H20minus (once we support triatomic molecules)
- I also added code for Nminus and Cminus, but a comparison against another source makes me a little worried about these. I've disabled them for now...


ajwheeler:
(Let me know if you want me to pick this up from here, happy for you to do it if you prefer)

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+        "H20" => [  996.0,    342.0,     28.0]      #-----     -----      -----      -----
+    )
+
+    Dict(k => LinearInterpolation(view(T_vals, 1:length(v)), v.*1e-34, extrapolation_bc = Throw())
+         for (k,v) in table)
+end
+
+
+# load free-free absorption data from John 1975, MNRAS, 172, 305. This data is valid for
+# 0.5 μm ≤ λ ≤ 10 μm. Most of these tables share a common set of λ and T values.
```
ajwheeler, line 75:
```suggestion
# Most John1975b tables share a common set of λ and T values.
```

The comment starting with "loads..." was confusing to me when I was trying to understand the structure of this file.

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+                        view(elements,size(elements)[1]:-1:1,:), extrapolation_bc=0)
+end
+
+const _short_wavelength_Heminus_ff_interp = let
+    # this comes from Table I of John 1975, MNRAS, 172, 305 and it holds He⁻ ff absorption
+    # cross sections (including the correction for stimulated emission)
+    #
+    # we don't plan to use this, particular data (we have more modern data for He⁻ ff absorption),
+    # this is primarily for testing
+    elements = SMatrix{7,9}(
```
ajwheeler, line 91: Given the discussion about stack space we had, do you think we should use static arrays here?  (I genuinely don't know.)

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+    LinearInterpolation((reverse(λ_vals), T_vals),
+                        view(elements,size(elements)[1]:-1:1,:), extrapolation_bc=0)
+end
+
+const _short_wavelength_Heminus_ff_interp = let
+    # this comes from Table I of John 1975, MNRAS, 172, 305 and it holds He⁻ ff absorption
+    # cross sections (including the correction for stimulated emission)
+    #
+    # we don't plan to use this, particular data (we have more modern data for He⁻ ff absorption),
+    # this is primarily for testing
```
ajwheeler, line 90:
```suggestion
```

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+const _typical_john75b_wavelengths = SA[1.0e-3, 5.0e-4, 2.5e-4, 1.5e-4, 1.0e-4, 7.5e-5, 5.0e-5] #cm
+const _typical_john75b_T_vals = SA[100.0, 500.0, 1000.0, 2500.0, 5000.0, 7500.0, 10000.0, 12500.0,
+                                   15000.0] #K
+
+function _build_interpolator_john75b(λ_vals, T_vals, elements)
+    LinearInterpolation((reverse(λ_vals), T_vals),
+                        view(elements,size(elements)[1]:-1:1,:), extrapolation_bc=0)
+end
+
+const _short_wavelength_Heminus_ff_interp = let
```
ajwheeler, line 85:
```suggestion

# for testing only!  We have a better He- ff treatment.
const _short_wavelength_Heminus_ff_interp = let
```

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+For a description of the kwargs, see [Continuum Absorption Kwargs](@ref).
+
+# Notes
+
+This function is defined for 100 K ≤ T ≤ 15000 K and 0.5 μ ≤ λ. It uses data from 
+[John 1975, MNRAS, 170, 5](https://ui.adsabs.harvard.edu/abs/1975MNRAS.170....5J) for λ > 10 μm
+and data from [John 1975, MNRAS, 172, 305](https://ui.adsabs.harvard.edu/abs/1975MNRAS.172..305J)
+for λ ≤ 10 μm. This is the approach recomended by the latter paper. The error on these absorption
+coefficients is probably large.
+"""
```
ajwheeler, line 202: I think this file should be restructured somehow to avoid the very similar doctrings.  Do functions for individual molecules need to exist?

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+                  3.60   4.52   6.54   8.61   1.06e1 1.42e1 1.71e1 2.15e1;
+                  3.12   3.96   5.80   7.69   9.56   1.27e1 1.52e1 1.88e1;
+                  2.79   3.57   5.28   7.05   8.80   1.15e1 1.37e1 1.70e1;
+                  2.29   2.97   4.40   5.87   7.27   9.77   1.18e1 1.45e1;
+                  1.98   2.57   3.85   5.15   6.42   8.70   1.06e1 1.31e1;
+                  1.75   2.29   3.45   4.63   5.79   7.88   9.62   1.20e1;
+                  1.58   2.08   3.14   4.22   5.29   7.25   8.88   1.12e1;
+                  1.45   1.91   2.90   3.92   4.92   6.77   8.31   1.05e1;
+                  1.22   1.62   2.49   3.39   4.29   5.95   7.35   9.33;
+                  1.07   1.43   2.21   3.04   3.86   5.38   6.67   8.50] .* 1e-27
```
ajwheeler, line 327: No need to change it, but I don't think you need the semicolons here.

test/neutral_metal_ff.jl
```diff
@@ -0,0 +1,60 @@
...
+                                                    nₑ)
+
+        # there's a factor of ~8 difference in the results. This seems a little extreme.
+        # Admittedly, John 1975 speculates that the error on their values at 10 μm might be a
+        # factor of 2. But, this still seems disturbingly big...
+        #
+        # A better test might use reference data from John & Williams 1977
+        @test assert_allclose(calc_α, ref_α; rtol = 6.8, atol =0.0);
+    end
+end
```
ajwheeler, line 60: Do the bell and berrington coefficients need tests?

ajwheeler (changes requested):

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
mabruzzo: The rules for array literals always confuse me

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+using Interpolations: LinearInterpolation, Throw
+
+# This file defines functions to compute the free-free linear absorption coefficients for
+# interactions involving neutral metals or neutral molecules. Based on convention, these
+# interactions are all named after negative ions.
+
+# -----------------------------------------------------------------------------------------
+# Define absorption functions by combining data from both
+# - [John 1975, MNRAS, 170, 5](https://ui.adsabs.harvard.edu/abs/1975MNRAS.170....5J)
+# - [John 1975, MNRAS, 172, 305](https://ui.adsabs.harvard.edu/abs/1975MNRAS.172..305J)
```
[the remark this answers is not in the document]
mabruzzo: Sure - I don't really care about whether we use markdown

And I agree about using John 1975a & John 1975b - I had that thought just before I pushed the PR, but was being lazy. I'll fix that

test/neutral_metal_ff.jl
mabruzzo: I was kind of trying to test `_bell_berrington_ff_neg_ion_absorption` here by comparing Nminus ff data from Ramsbottom et al. 1992 against Nminus ff data from John 75a (that could be better documented).

With that said, an obvious, more direct test is obvious to me for both `Nminus_ff` and `Cminus_ff`. I'll get to that in the next few days.

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
```diff
@@ -0,0 +1,422 @@
...
+# This is the procedure recommended by the latter paper.
+# -----------------------------------------------------------------------------------------
+
+"""
+    _combined_john_neg_ion_absorption(ν, T, ndens_neutral, ne, small_wavelength_cross_section,
+                                      large_wavelength_cross_section)
+
+uses absorption data from John 1975, MNRAS, 172, 305 for 0.5 μm ≤ λ ≤ 10 μm and data from
+John 1975, MNRAS, 170, 5 for λ > 10 μm. Returns linear absorption coefficient (in cm⁻¹)
```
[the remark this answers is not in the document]
mabruzzo: there's a typo in your suggestion, I'll add this manually

src/ContinuumAbsorption/absorption_ff_neutral_metals_molecules.jl
mabruzzo: Combining these seems sensible to me. I'm inclined to leave the metals as separate functions since they all use data from different papers, but if you feel super strongly about this, I can probably be convinced

ajwheeler: That seems sensible to me.  The tabulated data are in slightly different formats, correct?

mabruzzo: Ominus_ff uses data from John 75a and John 75b. 
- The format of John75a is pretty different from the formats of everything else - to make these quantities equivalent to the other sources, we just interpolate in T and then multiply the result by λ^2 (in units of Angstroms^2). This "common format" just requires that the computed "coefficient" is multiplied by the partial electron pressure and the number-density of the relevant neutral species.
- John 75a involves interpolating over λ and T

Even though they come from different papers, from each other the Cminus_ff & Nminus_ff data share the same format (Bell & Berrington were coauthors on both of those papers). This format is very similar to John 75b except they use θ instead of T (which obviously is an important distinction if we aren't interpolating in log-space).

So, if you felt it was important that we force everything into a single function, it's doable. It just doesn't seem like an obvious improvement (I really don't think we'll ever need to add any other negative atomic ion free-free absorption calculations beyond what we already have: `Hminus_ff`, `Heminus_ff`, `Cminus_ff`, `Nminus_ff`, and `Ominus_ff`)

ajwheeler: I‘ll leave this to your judgement.

mabruzzo:
At this point, I don't think I'll get back to this for a while. So the choice is totally up to you.

I don't really think most of these source are important - so it doesn't really matter if you merge before publishing. But, my impression was that there wasn't much work to do to get this merged in. With that in mind, and for the sake of selling the code to more naive readers (who are unaware that these sources aren't important), this might be worth doing since it adds a bunch of sources that turbospectrum has.

PR #113, opened 2022-05-26
Jet improvements

ajwheeler:
I used JET.jl and `@code_warntype` to identify places where type inference wasn't working as well as it could and fixed them.  Some of these were subtle, some were just me being an idiot about the type system.  This makes linelist parsing roughly twice as fast, and speeds up synthesis by ~15% for the APOGEE example (0.8s to 0.7s on my computer).

Contrary to what the commit message says, \alpha_cntm would not cause things to crash if not passed, but making it optional was tripping up the type system, so I made it required.


src/line_absorption.jl
```diff
@@ -24,8 +24,8 @@ is chosen.
 - if `α_cntm` is not passed, defaults to `window_size`, which is 2e-7 (in cm, i.e. 20 Å) unless
   otherwise specified
 """
-function line_absorption!(α, linelist, λs, temp, nₑ, n_densities::Dict, partition_fns::Dict, ξ
```
ajwheeler, line 27: α_cntm being optional was actually a bug.  If not passed it would crash.

src/linelist.jl
```diff
@@ -34,33 +34,30 @@ struct Formula
...
+    function Formula(code::String) 
         if code in atomic_symbols #quick-parse single elements
             return new([0, 0, atomic_numbers[code]]) 
         end
 
         #handle numeric codes, e.g. 0801 -> OH
         if all(isdigit(c) for c in code)
             if length(code) <= 2
-                return Formula(parse(Int,code))
+                return Formula(parse(Int,code))
```
mabruzzo, line 45:
```suggestion
                return Formula(parse(Int,code))
```

mabruzzo (approved):
Looks good to me. I suggested a minor whitespace tweak.

> speeds up synthesis by ~15% for the APOGEE example (0.8s to 0.7s on my computer).
Impressive!

PR #116, opened 2022-06-10
New atomic partition funcs

ajwheeler:
This introduces custom atomic partition funcs calculated from the NIST values up to 100,000K.  It includes a script to download the data, tabulate values, and stick them in an HDF5 file, which Korg pulls from.  It also "flat-extrapolates" the molecular equilibrium constants past 10,000K, and removes the filtering of atmospheric layers hotter than 10,000K.

This does address the problem of what to do about lowering the ionization potential / truncating energy levels / the occupation probability formalism.  I'm still mulling that over, but I think this is a good first step regardless.


src/linelist.jl
```diff
@@ -214,7 +213,7 @@ struct Line{F}
     end
 end
 
-#pretty-print lines in REPL and jupyter notebooks
+# it's important that this produces something parsable by the constructor
 function Base.show(io::IO, m::MIME"text/plain", line::Line)
     show(io, m, line.species)
```
mabruzzo, line 218:
```suggestion
function Base.show(io::IO, line::Line)
    show(io, line.species)
```

src/read_statmech_quantities.jl
```diff
@@ -36,9 +36,9 @@ the paper for more details.
 Note that the paper seems to suggest that we could actually use cubic interpolation. That should be
 revisited in the future.
```
mabruzzo, line 37:
```suggestion
```

mabruzzo (approved):

src/atmosphere.jl
```diff
@@ -90,7 +85,7 @@ function read_model_atmosphere(fname::AbstractString; truncate_at_10000K=true) :
 
         layers = map(zip(blockA, blockB)) do (lineA, lineB)
             nums = parse.(Float64, strip.(split(lineA * lineB)))
-            temp = nums[5]
+            temp = nums[5] # five digits doesn't let you go higher than our partition funcs top out.
```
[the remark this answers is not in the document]
ajwheeler: This was a thinko.  I've removed it.

PR #117, opened 2022-06-23
dumb bug in total_continuum_absorption

ajwheeler:


ajwheeler:
Nope, that's H2plus, but I just figured that out so I don't know what's going on yet.  I'll include the fix in this PR.

PR #123, opened 2022-09-19
New bezier-based formal solution

ajwheeler:
This introduces a new, optional, radiative transfer scheme alongside the old one.  It also switched to using quadrature for integrating over mu using the old method.  It would be nice to switch to this method as the default in the future, but there are complications which I am writing up in the paper.  It's currently a bit slower than the old solution, but I think it could be sped up to the same speed fairly easily.


src/synthesize.jl
```diff
@@ -131,8 +138,10 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
         α_cntm_layer = LinearInterpolation(cntmλs, α_cntm_vals)
         α[i, :] .= α_cntm_layer.(λs)
 
-        α5[i] = total_continuum_absorption([c_cgs/5e-5], layer.temp, layer.electron_number_density,
-                                           n_dict, partition_funcs)[1]
+        if ! bezier_radiative_transfer
+            α5[i] = total_continuum_absorption([c_cgs/5e-5], layer.temp, 
+                               layer.electron_number_density, n_dict, partition_funcs)[1]
```
mabruzzo, line 143:
```suggestion
                                               layer.electron_number_density, n_dict,
                                               partition_funcs)[1]
```
Minor whitespace nitpick

src/CubicSplines.jl
```diff
@@ -79,4 +79,34 @@ function (A::CubicSpline{<:AbstractVector{<:Number}})(t::Number)
     I + C + D
 end
 
+function cumulative_integral!(out, A::CubicSpline, t1, t2)
```
mabruzzo, line 82: I know this is mostly just a helper function, but could you add a small comment expressing your assumptions about the length of `out`?

mabruzzo (approved):
Looks great to me! I just have 2 incredibly minor comments.

(At some point, when I have more time, I want to go back and take a careful look at Bezier-based integration so that I have a better understanding of what's going on - but that's purely for my own edification)

PR #124, opened 2022-09-30
Refactored abundances specification

ajwheeler:
This factors the complicated arguments for specifying abundances  out of `synthesize` and into a new function `construct_abundances`.  It returns a vector of A(X) format abundances, which synthesize accepts.  I'd like feedback on the names of things (the new function and it's arguments).


src/synthesize.jl
```diff
@@ -85,39 +81,42 @@ function synthesize(atm::ModelAtmosphere, linelist, λ_start, λ_stop, λ_step=0
...
+    linelist = filter(l-> λs[1] - line_buffer*1e-8 <= l.wl <= λs[end] + line_buffer*1e-8, linelist)
+    
+    #check that λs is sorted
     if step(λs) < 0
         throw(ArgumentError("λs must be in increasing order."))
     end
 
-    #discard lines far from the wavelength range being synthesized
-    linelist = filter(l-> λs[1] - line_buffer*1e-8 <= l.wl <= λs[end] + line_buffer*1e-8, linelist)
+    if A_X[1] != 12
```
mabruzzo, line 108: Since this is the main API function, I think we should be explicit about the behavior of synthesize, when `A_X` has the "wrong" size.

If it holds just 1 element, then we currently assume that the star purely consists of hydrogen... That's fine, but I feel like we need to coerce the length of `A_X` or more importantly of `abs_abundances` (so we avoid other problems later on). Alternatively, we could just raise an error of `A_X` has the wrong length.

src/synthesize.jl
```diff
@@ -160,39 +159,74 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
...
+   than He. It is overriden by `abundances`.  
+- `abundances` is a `Dict` mapping atomic numbers or symbols to [``X``/H] abundances.  (Set 
+  `solar_relative=false` to use ``A(X)`` abundances instead.) These override `metallicity`.
+
+# Keyword arguments
+- `solar_relative` (default: true): When true, interpret abundances as being in \\[``X``/H\\] 
+  (``\\log_{10}`` solar-relative) format.  When false, interpret them as ``A(X)`` abundances, i.e. 
+   ``A(x) = \\log_{10}(n_X/n_\\mathrm{H}) + 12``, where ``n_X`` is the number density of ``X``.
+   Note that abundances not specified default to the solar value, adjusted with `metallicity`, in
+   either case.
```
mabruzzo, line 179: On the remote chance you didn't think deeply about this (while you were reorganizing), this keyword argument seems like it could potentially be removed (since the user can directly modify the output). 

With that said, I expect that you already thought about this and decided that it was important to included this option to make things more convenient to users (which I think is a totally valid choice) - if that's the case just resolve this conversation.

src/synthesize.jl
```diff
@@ -160,39 +159,74 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
...
-Calculate ``n_X/n_\\mathrm{total}`` for each element X given abundances in either [``X``/H] or 
-``A(X)`` form.  See [`synthesize`](@ref) for a detailed description of the arguments. Returns a 
-vector indexed by atomic number.
+Returns a 92 element vector containing abundances in ``A(X)`` (``\\log(X/H) + 12``) format for 
+elements from hydrogen to uranium.
+
+# Arguments
+You can provide either or both of:
+- `metallicity` (default: 0), i.e. [metals/H] is the ``\\log_{10}`` solar-relative abundance of elements heavier 
+   than He. It is overriden by `abundances`.
```
mabruzzo, line 170:
```suggestion
   than He. It is overriden by `abundances`, on a per-element basis.  
```

Not strictly necessary, but I think it helps add clarity.

src/synthesize.jl
```diff
@@ -160,39 +159,74 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
...
-vector indexed by atomic number.
+Returns a 92 element vector containing abundances in ``A(X)`` (``\\log(X/H) + 12``) format for 
+elements from hydrogen to uranium.
+
+# Arguments
+You can provide either or both of:
+- `metallicity` (default: 0), i.e. [metals/H] is the ``\\log_{10}`` solar-relative abundance of elements heavier 
+   than He. It is overriden by `abundances`.  
+- `abundances` is a `Dict` mapping atomic numbers or symbols to [``X``/H] abundances.  (Set 
+  `solar_relative=false` to use ``A(X)`` abundances instead.) These override `metallicity`.
```
mabruzzo, line 172:
```suggestion
  `solar_relative=false` to use ``A(X)`` abundances instead.) These override `metallicity`.
  This is the only way to specify an abundance of He that is non-solar.
```
This is probably not necessary, but I think it could be useful. 

This is mostly motivated by the fact that I initially found the fact that metallicity doesn't affect He abundances to be somewhat non-obvious. After thinking about it, I realize that it makes a LOT of sense, but I think there might be some values in making things obvious for new users (especially for new students who might not have this intuition).

src/synthesize.jl
```diff
@@ -160,39 +159,74 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
 end
 
 """
-    get_absolute_abundances(metallicity, abundances, solar_abundances, solar_relative)
+    construct_abundances(metallicity, abundances)
```
mabruzzo, line 162:
```suggestion
    construct_abundances(metallicity, abundances; kwargs... )
```

mabruzzo (approved):
I think this looks fantastic. I have a few minor suggestions, but they're all very style based (feel free to ignore anything).

It would definitely be worth having @andycasey weigh-in since he actually uses the code and can provide more valuable insight into the API's ergonomics than me.

A few incredibly minor general comments
- Ideally, `construct_abundances` might have a slightly more descriptive name (describing the exact format of abundances). But, I don't think that's possible without using an unweildy name.
- Since `synthesize` is the main API function, it seems like it might be helpful to perform more careful error checking for the arguments (e.g. make sure `A_X` has the right length or be explicit about what happens if the user gives the wrong length).


As an aside, it might eventually be helpful to explicitly document how a user might use auto-differentiation with relative abundances. I imagine, that the best way to do that is to have a page of documentation that walks people through doing this on the website (even if it's in the tutorial, some people don't really look at tutorial notebooks very much - especially after they feel like they have a basic understanding of what is going on). But this would need to be dealt with in an entirely separate PR.

src/synthesize.jl
ajwheeler: I've pushed a change to explicitly check that length(A_X) is right.  I only intended to support complete abundance specifications.

src/synthesize.jl
ajwheeler: This is actually not true.  You can override He  with the `abundances` dict regardless of whether with solar_relative is true or false, e.g. `construct_abundances(Dict("He"=>-1))` specified solar except with [He/H] = -1.

ajwheeler:
Thanks for the excellent feedback.  I've added a check in `synthesize` requiring that `A_X` have length 92.  I'm not really happy with the name `construct_abundances`, because I want something that is more explicit but also not super long.  Other options:
- `format_A_X`
- `format_abundance_vect` (blah)
- `construct_A_X`

src/synthesize.jl
mabruzzo: I think there might be a miscommunication here - my suggestion was just to add some text emphasizing that if you want a non-solar abundance of Helium, you need to specify it in the `abundances` dict (as I currently understand it, metallicity has no effect on Helium). This edit has nothing to do with the value of the `solar_relative` keyword argument.

I was suggesting we make a change equivalent to adding the italicized text in the following snippet:

> **Arguments**
> You can provide either or both of:
> - `metallicity` (default: 0), i.e. [metals/H] is the $log_{10}$ solar-relative abundance of elements heavier 
>    than He. It is overriden by `abundances`, on a per-element basis.  
> - `abundances` is a `Dict` mapping atomic numbers or symbols to $[X/H]$ abundances.  (Set 
>   `solar_relative=false` to use ``A(X)`` abundances instead.) These override `metallicity`.
>    _This is the only way to specify an abundance of He that is non-solar._

As I mentioned before, this is only a minor suggestion to help people like me (i.e. new students), who might be a little surprised that the metallicity argument has no bearing on the value of Helium. I totally understand if you don't think this is a worthwhile addition to make (I'm personally not convinced that it's particularly useful).

ajwheeler: Ah, my bad.  Sorry for the misunderstanding--I read the suggested change too quickly.

mabruzzo:
> Thanks for the excellent feedback. I've added a check in `synthesize` requiring that `A_X` have length 92. I'm not really happy with the name `construct_abundances`, because I want something that is more explicit but also not super long. Other options:
> 
>     * `format_A_X`
> 
>     * `format_abundance_vect` (blah)
> 
>     * `construct_A_X`

`make_abundance_vec` or `build_abundance_vec` are 1-2 characters shorter than `construct_abundances` & `vec` is an [idiomatic abbreviation of vector](https://docs.julialang.org/en/v1/base/arrays/#Base.vec) in Julia (I expect you won't like these).

My tentative vote is not to include `A_X` in the name  (I like `construct_abundances` more than `construct_A_X` or `format_A_X`), but I realize that's my own bias as someone outside this field

ajwheeler:
I'm a little hesitant about including "vec" in the function name as it makes the API a bit brittle (i.e. if we change the type of the object returned we would have to change the method name).  I'm not strongly against it, though.  

Would something like `abs_abundance_vec`  check your boxes in terms of explicitness?

src/synthesize.jl
```diff
@@ -160,39 +159,75 @@ function synthesize(atm::ModelAtmosphere, linelist, λs::AbstractRange; metallic
 end
 
 """
-    get_absolute_abundances(metallicity, abundances, solar_abundances, solar_relative)
+    construct_abundances(metallicity, abundances; kwargs... )
 
-Calculate ``n_X/n_\\mathrm{total}`` for each element X given abundances in either [``X``/H] or 
-``A(X)`` form.  See [`synthesize`](@ref) for a detailed description of the arguments. Returns a 
-vector indexed by atomic number.
+Returns a 92 element vector containing abundances in ``A(X)`` (``\\log(X/H) + 12``) format for
```
mabruzzo, line 164:
```suggestion
Returns a 92 element vector containing abundances in ``A(X)`` (``\\log_{10}(X/H) + 12``) format for
```

mabruzzo:
> I'm a little hesitant about including "vec" in the function name as it makes the API a bit brittle (i.e. if we change the type of the object returned we would have to change the method name). I'm not strongly against it, though.

That's an excellent point. You know I've come full circle and I'm fine with `construct_abundances`.

I think I might like `format_abundances` marginally more since the function coerces the user-specified abundance information to our preferred "format"
- for specifying abundances: $A(X) = \log_{10}(X/H) + 12$
- of organizing the abundances in memory (currently, a vector)

> Would something like `abs_abundance_vec` check your boxes in terms of explicitness?

Even with the caveat that you mentioned, I also like this name. 

At the end of the day, this is all just bikeshedding, and I it doesn't really matter which of these names (or the other names you suggested) are probably fine.

ajwheeler:
Thanks for the help with the documentation.  I'll ask Andy for comment on the function name in our meeting tomorrow then pick something and merge.

ajwheeler:
Recording for the record: Andy and I both like `format_A_X`.

PR #134, opened 2022-11-09
LSF speed and accuracy

ajwheeler:
This is not a replacement for truly fast LSF calculation, but it's a slight improvement.  It also fixes a bug in the LSF code.  Where I was smoothing only with a 1sigma wide kernel before (no idea wtf I was thinking), now I am smoothing with a 3sigma wide kernel.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -57,7 +57,7 @@ function constant_R_LSF(flux::AbstractVector{F}, wls, R) where F <: Real
     for i in 1:length(wls)
         λ0 = wls[i]
         σ = λ0 / R / 2
-        lb, ub = move_bounds(wls, lb, ub, λ0, σ)
+        lb, ub = move_bounds(wls, lb, ub, λ0, 3σ)
         ϕ = normal_pdf.(wls[lb:ub] .- λ0, σ)
         normalization_factor[i] = 1 ./ sum(ϕ)
         @. convF[lb:ub] += flux[i]*ϕ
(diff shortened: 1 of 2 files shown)


andycasey (approved):
The 1σ likely explains some of the earlier things that Sarah was seeing. 3σ should be enough for now but at some point we might want to make this an optional argument.

ajwheeler:
No reason not to do that now, honestly.  I really can't get my head around what on earth I was thinking.

mabruzzo (approved):
Looks good to me

src/utils.jl
```diff
@@ -49,18 +50,18 @@ the region you are going to compare.
...
 """
-function constant_R_LSF(flux::AbstractVector{F}, wls, R) where F <: Real
+function constant_R_LSF(flux::AbstractVector{F}, wls, R; window_size=3) where F <: Real
     #ideas - require wls to be a range object? Use erf to account for grid edges?
     convF = zeros(F, length(flux))
     normalization_factor = Vector{F}(undef, length(flux))
     lb, ub = 1,1 #initialize window bounds
     for i in 1:length(wls)
         λ0 = wls[i]
         σ = λ0 / R / 2
```
segasai, line 60: Shouldn't this be lamdbda/R/2.35 (or 2sqrt(2ln(2)))

andycasey: As usual, @segasai is correct. 

I’m embarrassed I missed this, particularly since — on the same day I reviewed this — I implemented the LSF convolution and resampling with a sparse matrix in Python, and there I used the correct 1/2.35 scaling. 🙃

ajwheeler:
Thanks, both.  I'm not sure how I've made it this far without knowing
this.  I'll open a PR.

PR #133, opened 2022-11-08
tweak fixed widths for parsing MARCS atmospheres

ajwheeler:

PR #136, opened 2022-11-11
make autodiff through atmosphere interpolation work (for now)

ajwheeler:
This generalizes some types, and addresses some inf/nan issues (see https://juliadiff.org/ForwardDiff.jl/stable/user/advanced/#Fixing-NaN/Inf-Issues) in the metal bf code.

It also precomputes n/U for each species in the linelist, rather than doing it for each line.

#133 should be merged first.

### DIFF
--- a/src/ContinuumAbsorption/absorption_metals_bf.jl
+++ b/src/ContinuumAbsorption/absorption_metals_bf.jl
@@ -36,11 +36,11 @@ function metal_bf_absorption!(α, νs, T, number_densities)
             end
             # cross-section interpolator
             σ_itp = metal_bf_cross_sections[spec]
-            σ = σ_itp.(νs, log10(T))
-            # Protecting this is necessary to avoid NaNs in derivatives.  
-            # It's probably better to avoid them in the first place for performance reasons.
-            mask = isfinite.(σ)
-            α[mask] .+= exp.(log(number_densities[spec]) .+ σ[mask]) * 1e-18 #convert to cm^2 
+            log_σ = σ_itp.(νs, log10(T))
+            # Using this mask is necessary to avoid NaNs in derivatives, which arise when σ = 0 and 
+            # log(σ) = -∞.
+            mask = isfinite.(log_σ)
+            α[mask] .+= exp.(log(number_densities[spec]) .+ log_σ[mask]) * 1e-18 #convert to cm^2 
         end
     end
 end
--- a/src/ContinuumAbsorption/absorption_metals_bf.jl
+++ b/src/ContinuumAbsorption/absorption_metals_bf.jl
@@ -36,8 +36,11 @@ function metal_bf_absorption!(α, νs, T, number_densities)
             end
             # cross-section interpolator
             σ_itp = metal_bf_cross_sections[spec]
-            α .+= exp.(log(number_densities[spec]) .+ σ_itp.(νs, log10(T))) * 1e-18 #convert to cm^2 
+            σ = σ_itp.(νs, log10(T))
+            # Protecting this is necessary to avoid NaNs in derivatives.  
+            # It's probably better to avoid them in the first place for performance reasons.
+            mask = isfinite.(σ)
+            α[mask] .+= exp.(log(number_densities[spec]) .+ σ[mask]) * 1e-18 #convert to cm^2 
         end
     end
-
 end
--- a/src/atmosphere.jl
+++ b/src/atmosphere.jl
@@ -90,7 +90,7 @@ function read_model_atmosphere(fname::AbstractString) :: ModelAtmosphere
             depth = parse(Float64, line[19:28])
             temp = parse(Float64, line[30:36])
             Pₑ = parse(Float64, line[39:48])
-            P = parse(Float64, line[51:60])
+            P = parse(Float64, line[50:60])
 
             nₑ = Pₑ / (temp*kboltz_cgs) # electron number density
             n = P / (temp*kboltz_cgs)   # non-electron number density
--- a/src/atmosphere.jl
+++ b/src/atmosphere.jl
@@ -58,35 +58,35 @@ end
 This is a convienince functions for making plots, etc.  Note that it doesn't access quantities in a 
 memory-efficient order.
 """
-tau_5000s(atm::ModelAtmosphere) = [l.tau_5000 for l in atm.layers]
+get_tau_5000s(atm::ModelAtmosphere) = [l.tau_5000 for l in atm.layers]
 """
    zs(atm::ModelAtmosphere) = [l.z for l in atm.layers]
 
 This is a convienince functions for making plots, etc.  Note that it doesn't access quantities in a 
 memory-efficient order.
 """
-zs(atm::ModelAtmosphere) = [l.z for l in atm.layers]
+get_zs(atm::ModelAtmosphere) = [l.z for l in atm.layers]
 """
    temps(atm::ModelAtmosphere) = [l.temp for l in atm.layers]
 
 This is a convienince functions for making plots, etc.  Note that it doesn't access quantities in a 
 memory-efficient order.
 """
-temps(atm::ModelAtmosphere) = [l.temp for l in atm.layers]
+get_temps(atm::ModelAtmosphere) = [l.temp for l in atm.layers]
 """
    electron_number_densities(atm::ModelAtmosphere) = [l.electron_number_density for l in atm.layers]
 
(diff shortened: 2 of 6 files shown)


src/synthesize.jl
```diff
@@ -152,7 +152,8 @@ function synthesize(atm::ModelAtmosphere, linelist, A_X::Vector{<:Real}, λs::Ab
     end
     #put number densities in a dict of vectors, rather than a vector of dicts.
     n_dicts = first.(pairs)
-    number_densities = Dict([spec=>[n[spec] for n in n_dicts] for spec in keys(n_dicts[1])])
+    number_densities = Dict([spec=>[n[spec] for n in n_dicts] for spec in keys(n_dicts[1]) 
+                             if spec != species"H III"])
```
mabruzzo, line 156: This may not necessarily be the time to make this change, but should the species constructor throw an error if a user tries to specify an ion that corresponds to charge that exceeds the atomic number?

src/line_absorption.jl
```diff
@@ -47,11 +54,12 @@ function line_absorption!(α, linelist, λs, temp, nₑ, n_densities, partition_
         γ = @. Γ * line.wl^2 / c_cgs
 
         E_upper = line.E_lower + c_cgs * hplanck_eV / line.wl 
-        levels_factor = (@. (exp(-β*line.E_lower) - exp(-β*E_upper)) /
-                             partition_fns[line.species](log(temp)))
+        #levels_factor = (@. (exp(-β*line.E_lower) - exp(-β*E_upper)) / partition_fns[line.species](log(temp)))
```
mabruzzo, line 57: Do we need to keep the commented line?

src/line_absorption.jl
```diff
@@ -47,11 +54,12 @@ function line_absorption!(α, linelist, λs, temp, nₑ, n_densities, partition_
...
+        levels_factor = (@. (exp(-β*line.E_lower) - exp(-β*E_upper)))
 
         #total wl-integrated absorption coefficient
-        amplitude = @. 10.0^line.log_gf*n_densities[line.species]*sigma_line(line.wl)*levels_factor
+        #amplitude = @. 10.0^line.log_gf*n_densities[line.species]*sigma_line(line.wl)*levels_factor
```
mabruzzo, line 61: Same as above (do we need to keep the commented line?)

src/line_absorption.jl
```diff
@@ -30,6 +31,12 @@ function line_absorption!(α, linelist, λs, temp, nₑ, n_densities, partition_
     lb = 1
     ub = 1
     β = @. 1/(kboltz_eV * temp)
+
+    # precompute number density / partition function for each species in the linelist
+    n_div_Z = map(collect(Set([l.species for l in linelist]))) do spec
+        spec => @. (n_densities[spec] / partition_fns[spec](log(temp)))
+    end |> Dict
```
mabruzzo, line 38: Do we actually save much time from computing this ahead of time? (if so, is it related to hidden auto-diff calculations?)

src/atmosphere.jl
```diff
@@ -51,6 +51,43 @@ function Base.show(io::IO, m::MIME"text/plain", atm::A) where A <: ModelAtmosphe
...
+"""
+electron_number_densities(atm::ModelAtmosphere) = [l.electron_number_density for l in atm.layers]
+"""
+    number_densities(atm::ModelAtmosphere) = [l.number_density for l in atm.layers]
+
+This is a convienince functions for making plots, etc.  Note that it doesn't access quantities in a 
+memory-efficient order.
+"""
+number_densities(atm::ModelAtmosphere) = [l.number_density for l in atm.layers]
+
```
mabruzzo, line 90: One thought that may be worth considering for a future PR. What if instead of making 5 helper functions that each handled one quantity, you made a single function that looked something like the following:

```julia
function atm_to_dict(atm::ModelAtmosphere)
    map([:tau_5000, :z, :temp, :electron_number_density, :number_density]) do prop
        String(prop) => [getfield(l,prop) for l in atm.layers] 
    end |> Dict
end
```

Since it's just a convenience function it's okay for this to be a little slow. Plus, you would be able to avoid adding these generic function names to namespace.

With that said, if you prefer your current solution, that's fine with me.

src/ContinuumAbsorption/absorption_metals_bf.jl
```diff
@@ -36,8 +36,11 @@ function metal_bf_absorption!(α, νs, T, number_densities)
             end
             # cross-section interpolator
             σ_itp = metal_bf_cross_sections[spec]
-            α .+= exp.(log(number_densities[spec]) .+ σ_itp.(νs, log10(T))) * 1e-18 #convert to cm^2 
+            σ = σ_itp.(νs, log10(T))
+            # Protecting this is necessary to avoid NaNs in derivatives.  
+            # It's probably better to avoid them in the first place for performance reasons.
+            mask = isfinite.(σ)
```
mabruzzo, line 42: what are the cases where σ is not finite? Outside of the interpolation table?

mabruzzo (approved):
This looks good to me. I basically just have a few minor clarifying questions.

src/atmosphere.jl
ajwheeler: I think this leaves a bit to be desired in just in that it requires thinking about a new object.  In general, I think that the singular form of a noun is almost always sensible as a variable name when the plural is.  I do actually get the hesitancy, and using `get_zs` type names might be a good alternative.

Btw, these changes were introduced in #133, but we can just merge this and delete that branch.

src/synthesize.jl
ajwheeler: That is a good idea.  I will add it to #1.

src/line_absorption.jl
ajwheeler: Thanks

src/line_absorption.jl
ajwheeler: Thanks

src/line_absorption.jl
ajwheeler: I tested this to make sure we saved time in normal cases.  It's not a breakthrough or anything but it is faster.  This was actually motivated by performance alone.  It's not necessary for autodiff to work.  Sorry, this is a bit of a frankenstein PR.

src/ContinuumAbsorption/absorption_metals_bf.jl
ajwheeler: Those are -Infs, which translate to 0s after being exp'd.  That process turned the derivative part of the dual numbers into NaNs.

ajwheeler: I realize that there are likely better ways to do this, but this produces correct results, and I don't think it's worth redoing unless this proves to be a speed bottleneck.

src/atmosphere.jl
ajwheeler: The more I ponder this, the more I'm leaning towards `get_temps` styles names.  Unless you object, I'm going to do that.

src/ContinuumAbsorption/absorption_metals_bf.jl
mabruzzo: That perfectly sensible to me!

It might be worth adding a line to the comments explaining where the NaNs come from (since the source non-obvious and presumably took you quite a bit of work to figure out)

src/atmosphere.jl
ajwheeler: I've done it.  Lmk if you hate it.

src/ContinuumAbsorption/absorption_metals_bf.jl
ajwheeler: done.

src/atmosphere.jl
mabruzzo: That's a good idea. I prefer that over what you currently have.

One last alternative (but again, feel free to ignore - I'm just bikeshedding here):
```julia
function query_atm_prop(atm::ModelAtmosphere, prop)
    prop_symbol = Symbol(prop)
    @assert prop_symbol in fieldnames(ModelAtmosphere)
    getfield.(atm.layers, prop_symbol)
end
```

ajwheeler: What is the reasoning for something like this over what I have?  Shorter implementation?

mabruzzo: Shorter implementation, fewer names in the namespace... It all comes down to personal preference. There's really no need to change it -- what you have definitely works. I'm going to hit merge

PR #137, opened 2022-11-11
Add bug report template

andycasey:
When I first used Julia I had no idea how to check which version I was using, or how to debug the stack trace. 

This template might make it easier for new Julia lovers to identify the versions of Julia and Korg that they're using when they create the report.


andycasey:
Hmm, it doesn't seem to automagically appear yet when I open a new issue. 

@ajwheeler: I think you might need to navigate to repository settings -> Issues -> Features -> Set up templates.

Sorry. I'd do it myself but I can't access the repository settings.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

Issue #140, opened 2022-11-12
Truncated Balmer lines


segasai:
Hi, 

Sorry for spamming in several issues.
This is related to #139 but in this case I looked at Balmer lines for teff=8000, logg=5.00, feh=-2.00 ,alpha/fe=0 vs PHOENIX.
And the Balmer lines are truncated. 
![compar1](https://user-images.githubusercontent.com/109513/201498465-7cbb86f5-78e0-4030-a279-7d0d19bbe37b.png)
If I had to guess there is some sort of window for the line that is used to compute the opacities.
Is it a bug or I need to change some parameter to fix that ? 

Thanks
        S

ajwheeler:
Thanks!  This is a bug in that the parameter you need to tweak isn't actually exposed to the user.  #141 fixes that, and makes the window size larger by default, so that you wouldn't need to tweak it in the first place for these inputs.

mabruzzo:
PR #141 fixed this, right? Can we close this issue?

ajwheeler:
I'm going to add an item to the big TODO issue so that automatic windowing doesn't drop off the radar forever, but I don't think it makes sense to work on now because it's not trivial and the hydrogen lines seem to be the part of the code most in need other changes.

PR #141, opened 2022-11-13
address Balmer line trunction

ajwheeler:
This addresses #140.  It makes the H line window size larger by default, and enables setting it when calling `synthesize`.

```julia
using Revise, Korg, PyPlot

atm = read_model_atmosphere("../../Korg_data/marcs_mod/p8000_g+5.0_m0.0_t01_ap_z-1.00_a+0.00_c+0.00_n+0.00_o+0.00_r+0.00_s+0.00.mod");

smallsol = synthesize(atm, [], format_A_X(), 4700, 5000; hydrogen_line_window_size=30)
bigsol = synthesize(atm, [], format_A_X(), 4700, 5000; hydrogen_line_window_size=150)
sol = synthesize(atm, [], format_A_X(), 4700, 5000)

plot(smallsol.wavelengths, smallsol.flux, label="small window")
plot(bigsol.wavelengths, bigsol.flux, label="big window")
plot(sol.wavelengths, sol.flux, "--", label="default")

legend()
xlabel("wavelength")
ylabel("flux")
```
image

### DIFF
--- a/src/line_absorption.jl
+++ b/src/line_absorption.jl
@@ -154,13 +154,11 @@ Arguments:
 - `ξ`: microturbulent velocity [cm/s]. This is only applied to Hα-Hγ.  Other hydrogen lines profiles 
    are dominated by stark broadening, and the stark broadened profiles are pre-convolved with a 
    doppler profile.
-- `stark_window_size`: the max distance from each line center [cm] at which to calculate the stark
-   broadening profile
-- `self_window_size`: the max distance from each line center [cm] at which to calculate the line 
+- `window_size`: the max distance from each line center [cm] at which to calculate the stark
+   and self broadening profiles
    absorption for Hα-Hγ (those dominated by self-broadening).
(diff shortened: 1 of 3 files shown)


segasai:
Thanks.
Should the window size to be determined semi-automatically based on the amount of absorption ? I.e. choose the window where the absorption is bigger than some threshold rather than hardcode the width.

ajwheeler:
That's how it works for everything except hydrogen lines.  I plan to use a similar treatment for them as well, but I want to at least make things workable by exposing the window size in the meantime.

ajwheeler:
The test that failed was comparing the code to old precomputed lyman series alphas.  
image


They are now ~7% apart at the wavelengths between the n=5 and n=6 lines.  I've changed the test to require a small atol instead (1e-8).

mabruzzo (approved):
Looks good to me.

Issue #142, opened 2022-11-14
NaNs in the spectra


segasai:
Hi, 

Continuing looking at the synthesized spectra I see that the hot star spectrum p8000_g+5.5_m0.0_t01_an_z-0.75_a-0.40_c+0.00_n+0.00_o-0.40_r+0.00_s+0.00.mod synthesized with this 

```
sol= synthesize(atm, lines, format_A_X(-0.75,Dict("O" => -1.150000,"Ne" => -1.150000,"Mg" => -1.150000,"Si" => -1.150000,"S" => -1.150000,"Ar" => -1.150000,"Ca" => -1.150000,"Ti" => -1.150000)), 3000, 10000);
```
shows a big chunk around 4000A filled with nans. See image
![image](https://user-images.githubusercontent.com/109513/201659481-9e5451ce-7389-4f89-a7fd-0a42223146b6.png)

(and somehow the chunk is now bigger than it was prior to the #141 fix.

Here's is the zoom on the region

![image](https://user-images.githubusercontent.com/109513/201660374-e7707b47-4315-4516-8af4-e1c844a9f3b0.png)
It looks the problem is centred at Hepsilon.

ajwheeler:
This is happening for parameters at which Stehlé & Hutcheon 1999 didn't tabulate values of the Stark-broadening profiles.  Working on a fix now.

PR #143, opened 2022-11-14
Fix nans in H lines

ajwheeler:
This addresses #142.

The change that fixes things is the new .h5 archive of the stark profiles, but I took the opportunity to slightly refactor.


mabruzzo (approved):
Looks great!

PR #144, opened 2022-11-15
correct LSF code to treat Δλ as the FWHM

ajwheeler:
Fixes the error pointed out by @segasai in #134.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -58,7 +58,7 @@ function constant_R_LSF(flux::AbstractVector{F}, wls, R; window_size=3) where F 
     lb, ub = 1,1 #initialize window bounds
     for i in 1:length(wls)
         λ0 = wls[i]
-        σ = λ0 / R / 2sqrt(2log(2)) # convert Δλ = λ0/R (FWHM) to sigma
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -57,7 +58,7 @@ function constant_R_LSF(flux::AbstractVector{F}, wls, R; window_size=3) where F
     lb, ub = 1,1 #initialize window bounds
     for i in 1:length(wls)
         λ0 = wls[i]
-        σ = λ0 / R / 2
+        σ = λ0 / R / 2sqrt(2log(2)) # convert Δλ = λ0/R (FWHM) to sigma
```
mabruzzo, line 61:
```suggestion
        σ = λ0 / R / (2sqrt(2log(2))) # convert Δλ = λ0/R (FWHM) to sigma
```
Shouldn't it actually be written this way? Even if it is accurate as written in Julia, I would still encourage you to make this change to make the logic easier to follow is easier for people coming from a different language

ajwheeler: Good call.  It is technically the same, but it's better to be unambiguous.

Issue #145, opened 2022-11-16
Scattered NaN values across spectra (continued)


segasai:
On my grid of MARCS models from time to time I'm seeing from 1 to 3 NaN values in the spectrum. I.e. for this one

```
  teff       flux            logg         mass   microt   feh    alpha                                        fname                                        abund_a abund_c abund_n abund_o abund_r abund_s                    synspec                    
float64    float64         float64      float64 float64 float64 float64                                      bytes82                                       float64 float64 float64 float64 float64 float64                    bytes49                    
------- ------------- ----------------- ------- ------- ------- ------- ---------------------------------------------------------------------------------- ------- ------- ------- ------- ------- ------- ----------------------------------------------
 4500.0 23252000000.0 5.500003068051694     0.0     0.0    0.75     0.4 MARCS/p4500_g+5.5_m0.0_t00_ae_z+0.75_a+0.40_c+0.00_n+0.00_o+0.40_r+0.00_s+0.00.mod     0.4     0.0     0.0     0.4     0.0     0.0 output/spec_4500_5.50_0.75_0.40_0.00_0.00.fits

```
Looking at the locations they seem to be single pixels at these 3 places 
3971.24, 4102.94, 9548.65  (sometimes just 1, sometimes 2 or 3).
Since those are Hydrogen line locations, presumably #143 is not complete fix.
Illustration of the problem (red pt shows a nan pixel)
![image](https://user-images.githubusercontent.com/109513/202174829-5338bff7-dc24-4ccd-8cc5-be540b827e8b.png)

ajwheeler:
Sorry about that.  I took another look at this and will open a PR momentarily which addresses the problem.  

I had been trying to avoid just flipping all the NaNs to 0s, but I've convinced myself that they will only arise in cases where that is the correct thing to do.  Furthermore, as long as I am interpolating the quantity I am (log absorption profile) and using Interpolations.jl, I'm going to have to flip the NaNs to 0s after the fact.

PR #146, opened 2022-11-16
Avoid NaNs in stark broadening profiles

ajwheeler:
addresses #145

### DIFF
--- a/src/line_absorption.jl
+++ b/src/line_absorption.jl
@@ -109,12 +109,16 @@ function _load_stark_profiles(fname)
             nes = read(fid[transition], "electron_number_densities")
             delta_nu_over_F0 = read(fid[transition], "delta_nu_over_F0")
             P = read(fid[transition], "profile")
+
+            logP = log.(P)
+            # clipping these to finite numbers avoid nans when interpolating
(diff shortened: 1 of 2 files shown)


src/line_absorption.jl
```diff
@@ -212,6 +210,8 @@ function hydrogen_line_absorption!(αs, λs, T, nₑ, nH_I, UH_I, ξ, window_siz
         ν₀ = c_cgs / (λ₀)
         scaled_Δν = _zero2epsilon.(abs.(view(νs,lb:ub) .- ν₀) ./ F0)
         dIdν = exp.(line.profile.(T, nₑ, log.(scaled_Δν)))
+        # interpolating between -Infs can lead to NaNs which should be 0
```
segasai, line 213: Isn't it better to clip the -inf to a -BIG_NUMBER instead ? 
Masking nans is indeed a bit worrysome.

segasai (approved):
Looks fine now. it's a bit unclear if the number -700 was chosen through some reasoning or not. Presumably to be low enough but  to not underflow to zero after exp-ing ?

ajwheeler:
Precisely, yes.  I've added a comment.

Issue #139, opened 2022-11-12
Comparison to PHOENIX in the CaT region


segasai:
Hi,

I've synthesized a bunch of spectra in the 3000A to 10000A wavelength range using Korg on a logg/teff/feh/alpha grid that I can compare with PHOENIX. One of the features I observe is major difference around the CaT (in the wings particularly). 
![compar](https://user-images.githubusercontent.com/109513/201495261-cc78d275-779f-4bd7-8931-c3846f059ae1.png)
here's a T=4500K, logg=5, [Fe/H]=0 [a/Fe]=0 case
I understand that you are not responsible for PHOENIX, but I heard a view from an expert )) that we see here is more likely korg's deficiency rather than PHOENIX due to some kind of strong line treatment. Do you have any advice/opinion on that ?  Or maybe I'm not using the right options ? 
 
Thanks 

PS I just used the standard Korg synthesize call with MARCS models + all the lines from VALD + individual alpha element abundances specified according to [Fe/H], [a/Fe].

ajwheeler:
> that we see here is more likely korg's deficiency rather than PHOENIX due to some kind of strong line treatment. Do you have any advice/opinion on that ?

My expectation is that this is correct, but I'm going to have to do some reading about PHOENIX's treatment of strong lines.  Thanks for bringing this to my attention.

segasai:
The PHOENIX paper says this on strong lines (sec 2.2 https://www.aanda.org/articles/aa/full_html/2013/05/aa19058-12/aa19058-12.html )  "The next step was to activate a mode in PHOENIX that triggers the use of special line profiles, which we used for the Ca lines, for example. In this mode no convergence is checked, so we carried out five more iterations by default in order to stratify the atmosphere again properly. Finally, the high resolution spectrum was synthesised from this intermediate model."
Pointed by @alexji

ajwheeler:
Thanks.  

I've been skimming the various PHOENIX papers, and I'm still unsure what they mean by "special" profiles.  Several papers mention that they use Doppler profiles for weak lines, and Voigt profiles for sufficiently strong lines, but it doesn't seem like that's what's being described here.

[...]

Issue #43, opened 2021-06-15
`voigt` could be more accurate (maybe OK?)


ajwheeler:
At present, we are using the approximation from Hunger 1965, which is the same at MOOG.  Previously, I thought there was a typo in the source, but it's looking more likely that the inaccuracies are just a limit of the approximation.  Unfortunately, this means that our voigt function breaks my 1% precision "rule".  At some point, we should look into doing better.

Note that in most of parameter space, where the Lorentz width is less than the gaussian width, there is no problem.  SO maybe this doesn't need to be fixed.  To put this to rest, I would at least like to verify that we aren't hitting that part of parameter space for any realistic problems.
![image](https://user-images.githubusercontent.com/711963/122097034-38d6cd80-cddd-11eb-8992-4c03441a9a18.png)

ajwheeler:
There are minor nonmonotonicities at $v = 5$. (discovered in #147)  For example:
![image](https://user-images.githubusercontent.com/711963/204603841-c2906359-02b2-4026-85b9-adf818c1c24b.png)

These are very unlikely to cause problems.

MFWStudy:
Hello,
I’m an undergrad student at Monash Uni doing some research over the summer break, and I’ve been looking into what sort of options we might have in improving the current Voigt approximation. 

I took a look through the literature, and I found three different approximations that looked like they might work well for Korg, all with fairly straightforward implementations:
-	Tepper-Garcia, 2006. https://ui.adsabs.harvard.edu/abs/2006MNRAS.369.2025T/abstract
-	The TurboSpectrum2019 (https://github.com/bertrandplez/Turbospectrum2019/blob/master/source-v19.1/newvoigt.f) implementation of Kuntz, 1997. https://ui.adsabs.harvard.edu/abs/1997JQSRT..57..819K/abstract
-	Hui, 1978. https://ui.adsabs.harvard.edu/abs/1978JQSRT..19..509H/abstract

I then tested all three of these, alongside the current Korg approximation, to see which were the fastest and most accurate. For the purposes of accuracy testing, I compared against the real part of the Faddeeva function found in Julia’s special functions library. 

[...]

PR #147, opened 2022-11-18
comments and refactoring in line profile code

ajwheeler:
#139 prompted me to revisit the line broadening code. It doesn't have any bugs* as far as I can tell, but I've added a bunch of comments to clarify some subtleties in how quantities are defined.  This is a work in progress, as I want to give the linelist parsing code the same makeover and refactor a bit more.

*technically the dynamic line windows are wrong, but only by order unity, which doesn't matter much.  I will fix this.

### DIFF
--- a/src/line_absorption.jl
+++ b/src/line_absorption.jl
@@ -33,18 +33,19 @@ function line_absorption!(α, linelist, λs, temp, nₑ, n_densities, partition_
     for line in linelist
         m = get_mass(line.species)
         
-        #doppler-broadening parameter
-        σ = doppler_width.(line.wl, temp, m, ξ)
+        # doppler-broadening width, √2 * σ
+        Δλ_D = doppler_width.(line.wl, temp, m, ξ)
 
-        #get all damping params from linelist.  There may be better sources for this.
+        # sum up the damping parameters.  These are FWHM (γ is usually the Lorentz HWHM) values in 
+        # angular, not cyclical frequency (ω, not ν).
         Γ = line.gamma_rad 
         if !ismolecule(line.species) 
             Γ = Γ .+ (nₑ .* scaled_stark.(line.gamma_stark, temp) +
                       n_densities[species"H_I"] .* [scaled_vdW(line.vdW, m, T) for T in temp])
         end
-        #calculate the lorentz broadenign parameter in in wavelength. Doing this involves an 
-        #implicit aproximation that λ(ν) is linear over the line window.
-        γ = @. Γ * line.wl^2 / c_cgs
+        # calculate the lorentz broadenign parameter in in wavelength. Doing this involves an 
+        # implicit aproximation that λ(ν) is linear over the line window.
+        Δλ_L = @. Γ * line.wl^2 / c_cgs
 
         E_upper = line.E_lower + c_cgs * hplanck_eV / line.wl 
         levels_factor = (@. (exp(-β*line.E_lower) - exp(-β*E_upper)) /
@@ -54,16 +55,16 @@ function line_absorption!(α, linelist, λs, temp, nₑ, n_densities, partition_
         amplitude = @. 10.0^line.log_gf*n_densities[line.species]*sigma_line(line.wl)*levels_factor
 
         ρ_crit = [cntm(line.wl) * cutoff_threshold for cntm in α_cntm] ./ amplitude
-        Δλ_D = maximum(inverse_gaussian_density.(ρ_crit, σ))
-        Δλ_L = maximum(inverse_lorentz_density.(ρ_crit, γ))
-        window_size = max(Δλ_D, Δλ_L)
+        dopper_line_window = maximum(inverse_gaussian_density.(ρ_crit, Δλ_D))
+        lorentz_line_window = maximum(inverse_lorentz_density.(ρ_crit, Δλ_L))
+        window_size = max(dopper_line_window, lorentz_line_window)
         lb, ub = move_bounds(λs, lb, ub, line.wl, window_size)
         if lb > ub
             continue
         end
 
-        invσ = 1.0./σ
-        @inbounds view(α, :, lb:ub) .+= line_profile.(line.wl, invσ, γ, amplitude, view(λs, lb:ub)')
+        invΔλ_D = 1.0./Δλ_D
+        @inbounds view(α, :, lb:ub) .+= line_profile.(line.wl, invΔλ_D, Δλ_L, amplitude, view(λs, lb:ub)')
     end
 end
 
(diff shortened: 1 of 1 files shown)


src/line_absorption.jl
```diff
@@ -33,18 +33,19 @@ function line_absorption!(α, linelist, λs, temp, nₑ, n_densities, partition_
...
+        # angular, not cyclical frequency (ω, not ν).
         Γ = line.gamma_rad 
         if !ismolecule(line.species) 
             Γ = Γ .+ (nₑ .* scaled_stark.(line.gamma_stark, temp) +
                       n_densities[species"H_I"] .* [scaled_vdW(line.vdW, m, T) for T in temp])
         end
-        #calculate the lorentz broadenign parameter in in wavelength. Doing this involves an 
-        #implicit aproximation that λ(ν) is linear over the line window.
-        γ = @. Γ * line.wl^2 / c_cgs
+        # calculate the lorentz broadenign parameter in in wavelength. Doing this involves an
```
segasai, line 46: broadenign -> broadening

ajwheeler: Thanks!

ajwheeler:
These changes rework the line absorption code to use the less standard, but more natural (to me) quantities.  This fixes the small line windowing bug, which arose because I had confused myself about what quantities I was using.  I've lowered the line cutoff threshold to 3e-4 times the continuum opacity (from 1e-3 times), since the more accurate line windows mean we can do so without sacrificing speed.

The changes to synthesized spectra are tiny:
image
image

Here, I am treating the microturbulence as a "width" in the "sqrt(2)sigma" sense, not the "sigma" sense (so it's definition is unchanged).  I think this is the normal definition.

ajwheeler:
btw, @mabruzzo this undoes some of the fine-tuning for speed that we did.  It is still slightly faster because of more accurate line windows, but we might be able to speed it up in future.

ajwheeler:
I think these tests are failing for reasons related to #43.

ajwheeler:
This PR is now ready for review.

The previously failing test is not the same as #43.  It's happening at $v := \frac{\lambda-\lambda_0}{\sqrt{2}\sigma} = 5$, which is near a different "seam" in our Voigt approximation.  I think this is acceptable, since it's a very small nonmonotonicity most of the time.  When it's large (meaning close to 10^-3 of the profile max) it will only be for one layer in the star.

Example:
image

segasai (approved):
It looks good to me -- I managed to understand what was going here, despite being a novice in ABO etc. 
The only thing that caught my eye now is the condition for the window to be max(doppler_window, lorentz_window). I understand there is no ideal solution here, but I'd probably either do sum's of windows or sqrt(sum of squares).

ajwheeler:
What is your thinking here? The amplitude of the voigt profile will be less than that of either of its components at any wavelength, so `max(doppler_window, lorentz_window)` should be an overestimate.

segasai:
Maybe I'm missing something here, but I thought that to zero order the window sizes under convolution should behave like the scale parameters of the distributions. I.e. when you convolve  gaussian/gaussian you get snew=sqrt(s1^2+s2^2) when you do the same with lorentz it's snew = s1+s2 so I thought the same would apply here.

ajwheeler:
The window_size variables are the detunings at which a pure Lorentz or pure Gaussian profile is 10^-3 x the continuum absorption.  The idea is to calculate the Voigt profile out to where is is negligibly small, taking its amplitude into account, so the window size won't scale directly with either width parameter.

segasai:
Okay, I can see that in the extreme case where the gaussian/lorentz windows are similar the window of voigt is ~ 25% larger than either.  

Python snippet:
```python
In [107]: def window(sig,gam, thresh=1e-3):
     ...:     def func(x):
     ...:        return scipy.special.voigt_profile(x,sig,gam)- thresh
     ...:     return scipy.optimize.fsolve(func,(sig+gam))
     ...: 

In [108]: window(6.17,0)
Out[108]: array([17.8164942])

In [109]: window(0,1)
Out[109]: array([17.81319416])

In [110]: window(6.17,1)
Out[110]: array([22.19586387])
```

ajwheeler:
Ah, good point.  I see the flaw in my reasoning now.  I'll update it to use the quadrature sum.

ajwheeler:
This actually resulted in a small speedup despite using slightly larger line windows, presumably because it avoids branching.  Thanks for catching my thinko!

src/line_absorption.jl
```diff
@@ -258,17 +271,15 @@ function sigma_line(λ::Real)
...
-    _line_profile(λ₀, invΔλ_D, Δλ_L*invΔλ_D/(4π), line_amplitude*invΔλ_D/sqrt(π), λ)
-end
-function _line_profile(λ₀::Real, invΔλ_D::Real, Δλ_L_invΔλ_D_div_4π::Real, 
-                       amplitude_invΔλ_D_div_sqrt_π::Real, λ::Real)
-    voigt(Δλ_L_invΔλ_D_div_4π, abs(λ-λ₀) * invΔλ_D) * amplitude_invΔλ_D_div_sqrt_π
+function line_profile(λ₀::Real, σ::Real, γ::Real, amplitude::Real, λ::Real)
+    inv_σsqrt2 = 1/(σ*sqrt(2))
+    scaling = inv_σsqrt2 / sqrt(π) * amplitude
+    voigt_hjerting(γ*inv_σsqrt2, abs(λ-λ₀)*inv_σsqrt2) * scaling
 end
```
mabruzzo, line 283:
```suggestion
function line_profile(λ₀::Real, σ::Real, γ::Real, amplitude::Real, λ::Real)
    inv_σsqrt2 = 1/(σ*sqrt(2))
    inv_sqrtπ = 1.0 / sqrt(π)
    scaling = inv_σsqrt2 * inv_sqrtπ * amplitude
    voigt_hjerting(γ*inv_σsqrt2, abs(λ-λ₀)*inv_σsqrt2) * scaling
end
```

Here is a minor optimization that shouldn't hurt readability and should convert divsion operations to multiplications.

As I recall, `Julia` is "smart" enough to precompute `inv_sqrtπ` when compiling `line_profile`. With that said, you might need to instead do something like the following (which would be necessary in just about any other language):
```julia
const inv_sqrtπ = 1.0 / sqrt(π)
function line_profile(λ₀::Real, σ::Real, γ::Real, amplitude::Real, λ::Real)
    inv_σsqrt2 = 1/(σ*sqrt(2))
    scaling = inv_σsqrt2 * inv_sqrtπ * amplitude
    voigt_hjerting(γ*inv_σsqrt2, abs(λ-λ₀)*inv_σsqrt2) * scaling
end
```

As an aside, the calculations for H-lines could probably be slightly optimized if `line_profile` function accepted `inv_σ` or `inv_σsqrt2` instead of `σ`. I'm sure you're well-aware of this, and I suspect this may not not be worth doing.

src/line_absorption.jl
```diff
@@ -219,15 +221,26 @@ function hydrogen_line_absorption!(αs, λs, T, nₑ, nH_I, UH_I, ξ, window_siz
...
 
-"the width of the doppler-broadening profile"
-doppler_width(λ₀, T, m, ξ) = λ₀ * sqrt(2kboltz_cgs*T / m + ξ^2) / c_cgs
+"""
+    doppler_width(λ₀ T, m, ξ)
+
+The standard deviation of of the doppler-broadening profile.  In standard spectroscopy texts, the 
+Doppler width often refers to σ√2, but this is σ
+"""
+doppler_width(λ₀, T, m, ξ) = λ₀ * sqrt(kboltz_cgs*T / m + (ξ^2)/2) / c_cgs
```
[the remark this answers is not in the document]
ajwheeler: Good catch.

src/line_absorption.jl
ajwheeler: >As an aside, the calculations for H-lines could probably be slightly optimized if line_profile function accepted inv_σ or inv_σsqrt2 instead of σ. I'm sure you're well-aware of this, and I suspect this may not not be worth doing.

I remember that is is true from when se optimized the previous code, but I don't remember why.

ajwheeler:
I have no idea why, but I just double checked, and both of these changed slowed down the code.  I'm going to undo them.  Let's return to it at speed week.

PR #150, opened 2022-11-23
Allow silly abundances if they match our silly prior

andycasey:
I hadn't run `grok` (the tool to compare Korg spectra with TurboSpectrum, MOOG, and SME) since the `format_A_X` function had been introduced.

My original `grok` wrapper for Korg failed because it said I wasn't allowed to specify an abundance of hydrogen -- because it's explicitly set as A(H) = 12.0 -- even if my abundance dictionary was setting A(H) = 12.0 as well. 

I think it's good to throw an error if someone tries to specify A(H) != 12 or [H/H] != 0, but my feeling is that Korg shouldn't throw an error if someone just provides `Dict("H"=>12)`.

If you disagree, feel free to close this PR without merging (and without explanation!).


src/synthesize.jl
```diff
@@ -202,7 +202,7 @@ You can provide either or both of:
 """
 function format_A_X(metallicity::Real=0.0, abundances::Dict=Dict();
                               solar_relative=true, solar_abundances=asplund_2020_solar_abundances)
-    if (1 in keys(abundances)) || ("H" in keys(abundances))
+    if (((1 in keys(abundances)) && (abundances[1] != 0)) || (("H" in keys(abundances)) && (abundances["H"] != 12)))
```
ajwheeler, line 205: Don't you also want to allow the cases where `abundances["H"]` is `0` or `abundances[1]` is `12` (depending on solar_relative)?  This could be moved down to `clean_abundances` is built up.  Then you wouldn't have to worry about whether the key is `"H"` or `1`.

src/synthesize.jl
```diff
@@ -216,13 +216,15 @@ function format_A_X(metallicity::Real=0.0, abundances::Dict=Dict();
                 throw(ArgumentError("$el isn't a valid atomic symbol."))
             elseif Korg.atomic_numbers[el] in keys(abundances)
                 throw(ArgumentError("The abundances of $el was specified by both atomic number and atomic symbol."))
+            elseif el != "H"
+                clean_abundances[Korg.atomic_numbers[el]] =  abund
             end
-            clean_abundances[Korg.atomic_numbers[el]] =  abund
```
ajwheeler, line 220: Is this change necessary?  (I think that it's not, at least if you make the change recommended above.)

src/synthesize.jl
```diff
@@ -216,13 +216,15 @@ function format_A_X(metallicity::Real=0.0, abundances::Dict=Dict();
...
-        elseif el isa AbstractString
-            if ! (1 < el < 92)
+        elseif el isa Integer
+            if ! (1 <= el < 92)
```
ajwheeler, line 223: Was this a bug?  Thanks.

src/synthesize.jl
```diff
@@ -202,7 +202,7 @@ You can provide either or both of:
 """
 function format_A_X(metallicity::Real=0.0, abundances::Dict=Dict();
```
ajwheeler, line 203: Do you mind adding a test case ensuring that the vector returned is the same regardless of of whether or not A(H) is specified?

ajwheeler (changes requested):

ajwheeler (approved):

ajwheeler:
I went ahead and made the changes I requested.  I'll review this and merge after I've forgotten it enough to do so.

PR #152, opened 2022-12-15
Interpolate marcs

ajwheeler:
This introduces code to automatically interpolated the SDSS MARCS atmosphere grid to whatever stellar parameters you like. The atmospheres are stored in an HDF5 file which is too big for github, which means we have to deal with it some other way. The plan is to store it in zenodo (could we use the SDSS server?).  

**option 1:** Make people manually download it and feed Korg its path.  
**option 2:**  When Korg is imported, check if `.korg/atmospherearchive.h5` exists, and automatically download it if it doesn't.  No idea how/if this would work on windows.
~~**option 3:** Don't use zenodo.  Pay for github LFS.  50GB of bandwidth per month is $5/month, which is very reasonable.  I'm happy to pay for this from my research budget, but if I stop doing for whatever reason it could be a snafu.  The other thing making this hard is that compiling Korg for CI would use all the bandwidth, so we would need a workaround for that.~~
**option 4:** Build a smaller grid into Korg, use option 1 if people want a better one.

Both options 2 and 3 have the problem that every time Korg is compiled by CI, the file gets re-downloaded.  I think we can work around this with an environmental variable, at least for option 2.  I think option 3 is the most technically elegant if the CI problems can be solved, but it has its own pitfalls.

TODO
- add tests (it's not clear that we can actually do this because of CI.  I could write a separate test suite to run on my personal computer.)
- resolve above quandary
- additional documentation (besides docstrings)?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,11 +5,13 @@ version = "0.12.1"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 FastGaussQuadrature = "442a2c76-b920-505d-bb47-c5924d526838"
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NLsolve = "2774e3e8-f4cf-5e23-947b-6d7e65073b56"
+ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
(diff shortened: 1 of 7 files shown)


ajwheeler:
In terms of performance, this works better than linearly interpolated synthesized spectra.  Here's some plots showing stddev and infnorm of the residuals vs synthesis with an exact atmosphere from 16000 to 16050 \AA convolved to apogee resolution. (Note that the scales are different.)
imageimage
imageimage

There's room for improvement on the cool end, but this is working well enough to ship, I think.

mabruzzo:
FWIW, I strongly think you should avoid option 3 at all costs.

The issues with option 3 are threefold:
1. Once the file is in the main branch's commit history it's basically there forever (I would strongly advise making edits to main branch's history - that could create all sorts of problems). We shouldn't count on fixing this later.
2. It will create issues with forking Korg (you would need to pay money to fork Korg). If Korg takes off in popularity (>10-20 contributors), people will eventually need to start forking Korg to make contributions (the current model of giving everybody write access to the repository won't scale well).
3. If you ever leave the field, hopefully Korg will live on after you. However, the need to pay money to keep the repositroy accessible will be an impediment to making that happen. (Furthermore, the monthly cost may be reasonable right now, but whose to say that won't change).

ajwheeler:
Those are good points. I am convinced that option 3 should be ruled out.

ajwheeler:
Ok, I've stuck the atmosphere archive in s3, which should cost < 0.01 cents/mo.  I've tested that this works but I'm not going to incorporate it into the automatic tests because we will run over what github will give us for free.

mabruzzo (approved):
This all looks good to me. The code for interpolation is a little hard to follow, but I'm not sure there's much that can be done.

PR #162, opened 2023-01-19
Fix issue introduced in 0.15.0

ajwheeler:
I introduced a problem in #155 which causes nans in autoderivatives.  This introduces a fix, as well as a test which will catch similar problems.  Unfortunately, this one wouldn't have been possible to catch easily since we can't test atmosphere interpolation in continuous integration.


andycasey (approved):
All looks good to me

PR #167, opened 2023-02-01
Un-beta-ify atmosphere interpolation

ajwheeler:
This is almost entirely a change in documentation, I'm providing more detail and updating the the tutorial.  The two behavior changes are the trying to download the atmosphere archive will throw an error if you have already done so (overridable), and that I'm printing the atmosphere loading message to stdout, rather than `@info`'ing it (the previous approach would be in an alarming red color in jupyter notebooks, and outputs the line number, which I think may confuse people.)

There is one thing that we should nail down before merging this.  The SDSS-IV website refers to the grid as being over [M/H], [alpha/M], [C/M], but the readme in the directory with the data refers to [Fe/H], [alpha/Fe], and [C/Fe].  We need to make sure our documentation and code use the correct parameters.


andycasey:
Holtz confirmed that the MARCS model atmospheres are [M/H], not [Fe/H].

ajwheeler:
I will push a fix.  This only affects documentation, argument names, and the `interpolate_marcs` method which takes `A_X` as an argument, which I don't think is being widely used.

ajwheeler:
Ready for review.  Please check my math in the `get_metals_H` and `get_alpha_H` if you don't mind.

andycasey:
I checked your math (and the rest), and it looks good to me.

PR #173, opened 2023-04-12
parse turbospectrum linelists

ajwheeler:
TODO 
- [x] iron out details wrt apogee linelist
- [x] tests
- [x] make faster?
- [x] remove old fw parsing code


ajwheeler:
I don't understand why this is failing on Ubuntu nightly, but I think it's fine to merge.  I also need to give the dead link checking a longer timeout, so the docs failing should also be ignored.

Issue #129, opened 2022-10-27
Korg fails for some atmospheres


ajwheeler:
**edit: [scroll down or click here for the current status](https://github.com/ajwheeler/Korg.jl/issues/129#issuecomment-1710641876)**

---

![image](https://user-images.githubusercontent.com/711963/198364690-aa4b0635-2ec4-4d88-bc08-19bae0925b21.png)

Details to come.  The cool stars failures are OK for now, since we are targeting FGK in paper 1, but obviously we should figure out what's going on.  I'm more concerned about the 3 in the middle.

ajwheeler:
The few points in the middle were just parsing problems. #130 fixes them.  

For the cool stars, the chemical equilibrium is either not converging, or returning negative number densities which causes things to crash later.  Two possibilities:
- the solver fails, because it's starting point is the analytically-calculated no-molecules solution, and cool stars have lots of molecules
- somehow things don't work because MARCS and Korg don't use the same set of species

segasai:
Hi, 
I don't quite know if it's the same issue as with cool stars but I also see failures for Teff=4000K MARCS
I.e. p4000_g+4.5_m0.0_t02_an_z+1.00_a-0.40_c+0.00_n+0.00_o-0.40_r+0.00_s+0.00.mod 
with this abundance (1.0, {'O': 0.6, 'Ne': 0.6, 'Mg': 0.6, 'Si': 0.6, 'S': 0.6, 'Ar': 0.6, 'Ca': 0.6, 'Ti': 0.6})


```
ERROR: DomainError with -3.50386098169805e6:
log will only return a complex result if called with a complex argument. Try log(Complex(x)).
Stacktrace:
  [1] throw_complex_domainerror(f::Symbol, x::Float64)
    @ Base.Math ./math.jl:33
  [2] _log(x::Float64, base::Val{:ℯ}, func::Symbol)
    @ Base.Math ./special/log.jl:301
  [3] log
    @ ./special/log.jl:267 [inlined]
  [4] metal_bf_absorption!(α::Vector{Float64}, νs::Vector{Float64}, T::Float64, number_densities::Dict{Korg.Species, Float64})
    @ Korg.ContinuumAbsorption ~/.julia/packages/Korg/Aas4k/src/ContinuumAbsorption/absorption_metals_bf.jl:39
[...]

PR #179, opened 2023-04-28
solve for ne instead of using atmospheric value

ajwheeler:
This branch is an experiment where Korg solves for the electron number density self-consistently, rather than assuming the model-atmospheric value is correct.  For the MARCS solar atmosphere, this adds about 0.1s to synthesis time, which is roughly double for a small wavelength range.

Notably, this implementation doesn't crash on any model atmospheres (at least at solar metallicity).  Contrast with the current method (#129), which sometimes fails for cool stars.

Here's the reletive difference in the $n_e$ that Korg infers using the old method and this one.  Solar atmosphere on the left, cool star (Teff = 3200K, logg=2.0 very close to the parameters which crash the current implementation) on the right:
imageimage

The agreement between the methods is worse for the cool star, which is consistent with Korg's solver failing near there.  Completely unexpectedly (to me), the new method actually produces closer agreement with the model atmosphere pretty much everywhere. I'd really like to know what the cause for disagreement with the atmosphere is, but my guess is charged molecules.

The seems like a slam dunk except that it's slow.  

TODO 
- [x] actually use new ne for continuum stuff, etc
- [x] make fast?
- [x] document changes in atmosphere code
- [x] add warning when electron number density diverges from model atmosphere
- [x] stress test to see where, if anywhere, Korg's chemical equilibrium can't converge

### DIFF
--- a/src/statmech.jl
+++ b/src/statmech.jl
@@ -98,20 +98,21 @@ Equilibrium constants are defined in terms of partial pressures, so e.g.
 function chemical_equilibrium(T, nₜ, model_atm_nₑ, absolute_abundances, ionization_energies, 
                               partition_fns, log_equilibrium_constants; x0=nothing)
     if x0 === nothing
-        x_type = promote_type(eltype(absolute_abundances), typeof(model_atm_nₑ), typeof(T), typeof(nₜ))
-        x0 = Vector{x_type}(undef, MAX_ATOMIC_NUMBER + 1)
         #compute good first guess by neglecting molecules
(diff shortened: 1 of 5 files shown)


ajwheeler:
OK, so I spent some time speeding this up and it's actually a bit faster than the old implementation now for cool stars where chemical equilibrium is more complicated (more molecules). 

new implementation: 70ms (sun),  148ms (cool giant)
old implementation: 70ms (sun) 227 ms (cool giant)

ajwheeler:
Github actions is currently down, but the tests pass on my computer.  This is ready for review.

PR #185, opened 2023-05-15
New Solar Abundance (Mass 2022)

CharleskingSpikeArphience:
The Z/X in the new data list is 0.02249704096881273, which agrees with Magg's paper.


ajwheeler:
Great! Could you rename the variable to `magg_2022_solar_abundances`?

PR #189, opened 2023-05-25
Custom forwarddiff rule

ajwheeler:
fast but experimental

related reading: 
- https://discourse.julialang.org/t/how-to-autodifferentiate-the-results-of-nlsolve/65680/2
- https://gist.github.com/YingboMa/c22dcf8239a62e01b27ac679dfe5d4c5


ajwheeler:
I wrote a wrapped version of `nlsolve`, but found that it wasn't adequate because I also really want to differentiate through the setup of the `residuals!` definition.  I'll stick with the custom method for `solve_chemical_equilibrium` and add guard rails to make sure it doesn't go wrong.

Wrapped `nlsolve` for posterity: https://gist.github.com/ajwheeler/36396be44e9d4b06c2e8a31552941b32

PR #197, opened 2023-06-29
Simple fit

ajwheeler:
I want to get working fitting code merged into main, so I'm scaling back my ambitions and trying to take things one at a time.  This has a subset of the functionality on the old branch.


ajwheeler:
Ok, this now has working, clean interface for fitting any combination of parameters to any subset of an observed spectrum.  It's basically ready to merge, but first I want to:
- look into speeding things up with precompilation.  First-time runs have a few minutes of compile time.
- make sure things are well-documented.
- write a tutorial notebook

Two other changes introduced here:
- There is now a custom exception type for when the chemical equilibrium solver fails.  This lets the fitting code catch those errors only.
- **`synthesize` now returns the continuum along with the flux** (@andycasey)

PR #200, opened 2023-07-18
Brackett series

ajwheeler:
Turns out these are important

### DIFF
--- a/src/line_absorption.jl
+++ b/src/line_absorption.jl
@@ -319,13 +319,20 @@ function hydrogen_oscillator_strength(n, m)
 end
 
 """
+    brackett_line_profile(m, λs, λ₀, T, nₑ, amplitude)
+
 Normalize stark-broadened line profile (specialized to Brackett series).  Translated and heavily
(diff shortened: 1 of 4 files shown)


ajwheeler:
This PR adds Brackett lines.  It treats them as broadened by the Stark effect only, which is a fairly accurate approximation.  This plot shows the half-width of the doppler, stark, and lorentz  profile for the 4->11 transition.  They are plotted as a function of optical depth for atmospheres across the SDSS grid in Teff, logg, and [M/H].  Other transitions are similar.
image

This makes Doppler broadening seem more important than it is, since Doppler profiles are so narrow. If I include the Doppler component ("convolving" in the same approximate way as Kurucz), the effect is only on line cores and is sub-1%.

ajwheeler:
(The failing tests are due to something with StaticArrays on Julia nightly, so they can be safely ignored. I'll let the pros sort that out.)

ajwheeler:
This is stated in the code, but it's worth mentioning here that this is a heavilly modified derivative of Kurucz's HLINOP (via Barklem: https://github.com/barklem/hlinop).

PR #205, opened 2023-07-21
Built in linelists

ajwheeler:
This makes a vald extract stellar linelist and the APOGEE linelist available without actually getting and parsing the files.


mabruzzo:
Out of curiosity how big are these files. (I imagine that you have a very good reason for adding these files - I'm just curious how large they are)

ajwheeler:
The big one is ~100Mg. I’m going to either handle it the same way as the
atmosphere archive, or repackage much smaller before squashing and merging.

PR #211, opened 2023-08-09
tweaks to abundance specification

ajwheeler:
This PR does 3 things
- It corrects the Grevesse 2007 solar abundances.  Not sure why they were wrong, but the PR in which they were introduced (#124) has no indication.  They were only used for calculating metallicity and alpha for the purposed of interpolating marcs atmospheres, so it's unlikely to have caused any science bugs for anyone.
- It includes C as an alpha element in `format_A_X` and `get_alpha_H`.  This is more consistent with MARCS and also more consistent the definition of the alphas (as I understand it).  Ofc people are always free to set each abundance to whatever they want.
- It changes the behavior of `get_metals_H` to ignore the alpha elements by default.  This is probably closer to what people intuitively want it to do, and it's more consistent with how MARCS defines things.  It also means that if you set alpha with `format_A_X`, it won't screw up your `get_metals_H` number.

@andycasey @andrew-saydjari  comments particularly welcome.  You don't need to check the code, just give thoughts on changes.


andycasey:
I think it’s sensible to follow the MARCS precedent.

Issue #202, opened 2023-07-20
Brackett line profiles, normalization, and level dissolution


ajwheeler:
In #200, I added support for the Brackett series based heavily on Barklem's [HLINOP.f](https://github.com/barklem/hlinop/blob/master/hlinop.f).  My `Korg.brackett_line_profile` and `STARK1` in `HLINOP.f` are doing the same calculations (though I have stripped out a lot as I only need to calculate profiles for Brackett lines).  I have confirmed by running the fortran directly that I am producing the same values, and Korg produces spectra with Brackett lines nearly identical to Turbospecrum (see below), which uses HLINOP. 

Here I show synthesized spectra in the APOGEE region (using the APOGEE linelist and model atmospheres) for a few sets of stellar parameters.  All have [M/H] = -1.5.  The last column shows the probability that the upper levels of the two zoomed-in lines are dissolved. (Tau=1 is marked with a vertical line.)

(Click to view bigger.)
![image](https://github.com/ajwheeler/Korg.jl/assets/711963/adeeb907-28a3-45fd-9805-3dc3b59848ae)

I think this shows clearly that in the high logg-cases, Korg and turbospectrum show less absorption because they are dissolving the upper levels of the transitions.  I was under the impression that synspec did the same, so it's worth tracking down the difference.

The Korg and Turbospectrum Brackett lines implementations are quite different from the synspec Brackett lines.  Synspec's implementation of the profiles is simpler, having only a Doppler core and an asymptotic Holtsmark wing.  The Korg/Turbospec implementation dynamically chooses how to model the core (Stark often dominates over Doppler), has a more sophisticated Holtsmark profile, and include the effect of impact broadening by electrons. 

<details>
<summary>Click to view the synspec stark profile source code</summary>

```fortran
C
C ********************************************************************
C
      FUNCTION STARKA(BETA,A,DIV,FAC)
C
C     Approximate expressions for the hydrogen Stark profile
C
C     Input: BETA  - delta lambda in beta units,
C            BETAD - Doppler width in beta units
C            A     - auxiliary parameter
C                    A=1.5*LOG(BETAD)-1.671
C            DIV   - only for A > 1; division point between Doppler
C                    and asymptotic Stark wing, expressed in units
C                    of betad.
C                    DIV = solution of equation
C                    exp(-(beta/betad)**2)/betad/sqrt(pi)=
C                     = 1.5*FAC*beta**-5/2
C                    (ie. the point where Doppler profile is equal to
[...]

PR #218, opened 2023-08-18
Brackett cleanup

ajwheeler:
This is a refactor of the brackett line code.  It also uses real convolution to deal with the 2 contributions to the stark broadening profile, which addresses #202.

This shows Brackett eta for HD84937.  Using the benchmark params gives a much better match than the APOGEE DR17 params (which came from fitting the wrong profile), indicating that the new profiles will yield less biased params.
image

The dashed line is included because the profiles adapted from HBOP are not perfectly normalized themselves. 
image

The question now is whether or not renormalize, though thankfully the impact on the spectrum is not huge.


ajwheeler:
This is now cleaned up to a state where I am happy with it.  Unfortunately, spending so long on the topic has left me unsatisfied with the theoretical backing.  I'm going to get this merged, but unfortunately I think it might be best to rewrite everything to use [Lemke 2009](https://ui.adsabs.harvard.edu/abs/1997A%26AS..122..285L/abstract) instead of the Griem papers that HLINOP and SYNTHE rely on.

PR #227, opened 2023-10-19
convert polyatomic parition funcs from physics to astrophysics convention

ajwheeler:
@andycasey can you sanity check this?


ajwheeler:
image
image

andycasey:
Looks much better. I didn't review the scraped data in detail, but the implementation looks right to me.

PR #239, opened 2023-11-29
fix bug preventing passing fixed_params as a dict when initial_guess …

ajwheeler:
…is also a dict

Thanks to @vedantchandra for reporting


ajwheeler:
This turned into a full refactor of `fit_spectrum` to use dicts internally, in the hopes that it would help compilation times.  It didn't, but the resulting code is a bit easier to follow, especially if you are not a Julia buff, so I'm keeping it.

PR #240, opened 2023-11-29
Allow LSFs with R as a function of wavelength

ajwheeler:
Does what it says in the title.  So far, I haven't changed the API at all, you just pass `R` as a function rather than a constant.  But I will have to deprecate `constant_R_LSF` because the name is now wrong.


ajwheeler:
`constant_R_LSF` has been renamed to `apply_LSF`. The old name is included as an alias but deprecated.

PR #242, opened 2023-12-05
comments, type hinting

ajwheeler:


ajwheeler:
Two other tiny edge cases fixed.

PR #246, opened 2024-01-02
better unify `compute_LSF` and `apply_LSF`

ajwheeler:
I would be nicer if these were more parallel still, but this is a step in the right direction that doesn't make breaking changes to the API.


mabruzzo (approved):
I only took a quick glance, but looks good to me!

PR #250, opened 2024-01-09
better/faster EWs-> abundances, and EWs -> stellar params

ajwheeler:
This PR significantly refactors `ews_to_abundances` and introduces a new function `ews_to_stellar_parameters`, which finds $T_\mathrm{eff}$, $\log g$, [Fe/H], and $v_\mathrm{mic}$ using the old-school method.

The speedup to `ews_to_abundances` is obtained by ensuring that it only calls synthesize once.  Rather than partitioning lines into different buffers so that their windows are non-overlapping, the new method chooses the boundary between each pair of too-close lines to be the wavelength with the least absorption.  I contend that this is reasonable, since if the EWs were measured in a real spectrum the lines can't be badly blended.  

The new function, `ews_to_stellar_abundances` uses a custom newton-raphson solver to find the parameters for which the slope of per-line abundances vs $\chi$, the slope of abundances vs REW, the difference between Fe I and Fe II, and the difference between the input and measured Fe I abundance are all 0.  It allows for custom max step sizes and tolerances for each parameter.

TODO
- [x] make it work with elements besides Fe?
- [x] warn the user when lines appear to be blended?
- [x] tests
- [ ] add parameter fitting to tutorial notebook
- [x] update documentation


ajwheeler:
In testing the changes to `ews_to_abundances`, I noted largest changes in inferred abundance were for lines in the wings of hydrogen lines.  Turning off hydrogen lines does this to the (raw/nondifferential) abundances from the Melendex EWs, which is a definite improvement.  
image

If I redo the differential analysis of all the Melendez elements, do slightly better than before.
old (top), new (bottom)
image
image

Also, I confirmed that increasing wavelength sampling tenfold changes inferred abundances only at the $10^-5$ level.  That's good to verify, and also indicate that it's not a problem that each line center is no longer exactly on a pixel value.

ajwheeler:
Note that #[3758efb](https://github.com/ajwheeler/Korg.jl/pull/250/commits/3758efb06e1d29e78e1bd908f030b9d587673bb5) changes the starting params `Teff0, logg0, vmic0, metallicity0` to kwargs.  Should we name these better/differently?

ajwheeler:
These final (fingers crossed) changes propagate EW uncertainties to uncertainties in inferred stellar parameters.  It is necessary to choose default uncertainties.  I'm currently assuming that all EWs have the same uncertainty be default. Note that homoscedastic EWs don't correspond _precisely_ to homoscedastic abundances, but it's very close.  This default behavior is thus very close to what you get if uncertainties are ignored at all steps.

Thinking about this as a procedure, and the fact that systematic uncertainties, rather than measurement uncertainties will often dominate the abundances, I decided to add an additional calculation.  
- "total error" (statistical and systematic) is calculated by assuming that all abundances have uncertainty equal to the scatter in abundances.
- This is propagated to get stellar parameter uncertainty, which are typically larger than the statistical ones.
- The systematic uncertainty is obtained by subtracting (in quadrature) the statistical uncertainty from the total.  It is set to zero when the total error is smaller, as could happen if the errorbars were inflated for some reason.
When no EW uncertainties are provided, all parameter uncertainty is reported as systematic.

src/fit.jl
```diff
@@ -406,16 +392,13 @@ A vector of abundances (`A(X) = log10(n_X/n_H) + 12` format) for each line in `l
 - `wl_step` (default: 0.01) is the resolution in Å at which to synthesize the spectrum around each 
    line. 
 - `ew_window_size` (default: 2): the farthest (in Å) to consider equivalent width contributions for 
-   each line.
+   each line.  It's very important that this is large enough to include each line entirely.
+- TODO blench_warn_threshold
```
andycasey, line 396: blend_warn_threshold docstring?

src/fit.jl
```diff
@@ -433,28 +416,241 @@ function ews_to_abundances(atm, linelist, A_X, measured_EWs, ew_window_size::Rea
...
+    if vmic0 == 0.0
+        throw(ArgumentError("Starting guess for vmic (vmic0) must be nonzero."))
+    end
+
+    # set up closure to compute residuals
+    get_residuals = (params) -> 
+        _stellar_param_equation_residuals(params, linelist, measured_EWs, measured_EW_err, 
+                                           fix_params, callback, passed_kwargs)
+
+    params = [Teff0, logg0, vmic0, metallicity0]
```
andycasey, line 552: I it sensible to clamp the initial parameters to be within `parameter_ranges` and show a warning if the initial value is outside the range? At the moment the behaviour is that an exception will be raised when trying to create an atmosphere. Maybe that's OK, but since there are explicit bounds for parameters, I would think it's sensible to clamp the initial parameters to be within the parameter bounds, and optionally show a warning.

src/fit.jl
```diff
@@ -433,28 +416,241 @@ function ews_to_abundances(atm, linelist, A_X, measured_EWs, ew_window_size::Rea
...
+
+
+    teff_residual = _get_slope([line.E_lower for line in linelist[neutrals]],
+                               A[neutrals], A_inv_var[neutrals])
+    logg_residual = (_weighted_mean(A[neutrals], A_inv_var[neutrals]) -
+                     _weighted_mean(A[.! neutrals], A_inv_var[.! neutrals]))
+    vmic_residual = _get_slope(REWs, A[neutrals], A_inv_var[neutrals])
+    feh_residual = _weighted_mean(A, A_inv_var) - (params[4] + Korg.grevesse_2007_solar_abundances[Z])
+    residuals = [teff_residual, logg_residual, vmic_residual, feh_residual]
+    residuals .*= .! fix_params # zero out residuals for fixed parameters
```
andycasey, line 592: Nice!

src/fit.jl
```diff
@@ -433,28 +416,241 @@ function ews_to_abundances(atm, linelist, A_X, measured_EWs, ew_window_size::Rea
...
+    total_ivar = ones(length(A)) * total_err^-2
+
+    stat_sigma, total_sigma = map([A_inv_var, total_ivar]) do ivar
+        sigma_mean = 1 ./ sqrt(sum(ivar))
+        teff_residual_sigma = _get_slope_uncertainty([line.E_lower for line in linelist[neutrals]], ivar[neutrals])
+        vmic_residual_sigma = _get_slope_uncertainty(REWs, ivar[neutrals])
+        [teff_residual_sigma, sigma_mean, vmic_residual_sigma, sigma_mean]
+    end
+
+    if EW_err == ones(length(EW))
```
andycasey, line 615: I still haven't fully digested how I feel about the systematic uncertainty estimates as we discussed, but I don't think it's wrong. I think it is worth including it in this merge and then thinking more about it.

andycasey (approved):
These are minor comments. I think the only functional suggestion change is to clamp the initial parameters to be within the parameter bounds. I'd like to test the uncertainty implementation in practice (after this is merged) before collapsing an opinion on it.

PR #251, opened 2024-01-16
first try at cache atmosphere archive in github actions

ajwheeler:
Here I'm attempting to cache Korg's atmosphere archive so that we can use in tests run by github actions without downloading it every time.

**warning: I will probably force-push to this branch to clean up the commit history before merging.**


ajwheeler:
Superseded by #252

PR #256, opened 2024-01-19
fix bug; add and refactor tests

ajwheeler:
Embarrassingly, #255 wasn't the only bug carelessly introduced by #252.

This addresses another one, adds a test that would have caught it, and refactors the atmosphere testing code generally.  There's now a function to facilitate testing approximate equality of model atmospheres, and atmosphere tests have been factored into a separate file.

### DIFF
--- a/src/atmosphere.jl
+++ b/src/atmosphere.jl
@@ -207,7 +207,7 @@ function interpolate_marcs(Teff, logg, A_X::AbstractVector{<:Real};
     C_H = A_X[6] - solar_abundances[6]
     C_M = C_H - M_H
     if clamp_abundances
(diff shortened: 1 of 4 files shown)


src/atmosphere.jl
```diff
@@ -207,7 +207,7 @@ function interpolate_marcs(Teff, logg, A_X::AbstractVector{<:Real};
     C_H = A_X[6] - solar_abundances[6]
     C_M = C_H - M_H
     if clamp_abundances
-        nodes = _sdss_marcs_atmospheres
+        nodes = _sdss_marcs_atmospheres[1]
```
ajwheeler, line 210: This is the bug

PR #257, opened 2024-01-30
refactor of radiative transfer, reëvaluation of numerical schemes

ajwheeler:
I am refactoring the radiative transfer code in order to retain the intensity information at all layers of the atmosphere.  This is 1) diagnostically useful, and 2) necessary to calculate the radiative pressure and net radiative flux.

This provides a natural opportunity to refactor another aspect of the transfer code.  When doing RT along a ray, there are two steps: integrating "away from the observer" to get the optical depths ($\tau$), and integrating back towards the observer to get the intensity ($I$). (It's a bit misleading to label these as "away"/"towards" the observer because you also need to account for inward-propagating rays to calculate the pressure, etc., but you get the idea.). At present, Korg has two numerical transfer integrators (detailed in Wheeler+ 2023) which inextricable link the schemes used for $\tau$ and $I$, but this need not be the case.

This (draft) PR decouples the $\tau$ and $I$ schemes, which provides more flexibility and also allows us to do some interesting tests.  It may be the case that after testing we go with an upgraded RT integrator with only one numerical scheme, provided that it meets performance and accuracy goals.

For testing, I'll use Teff=5000K, logg=3, [M/H]=0, with the built-in solar linelist and apogee linelist.  I'll do 300-500nm and 1600-16500nm to cover UV, visible and IR at a 0.001 nm resolution.  For reference, the whole synthesis takes 10.6s (dominated by hydrogen and water lines).

# performance 
current implementation times:
- default (archored + linear): 1.6 s
- bezier: 15 s

draft implementation times:
| | anchored $\tau$ (default) | bezier $\tau$ | 
| -- | -- | -- | 
| **linear $I$ (default)** | 2.9 s | 7.8 s |
| **bezier $I$** | 9.0 s | 13.7 s |

# accuracy 

Here, I'm comparing each variant to the default (which is likely itself the most accurate). First, let's look at the relative flux difference.  The comparison is very visible across most of the visible, and in the IR, the line cores are much more similar to the continuum, indicating that things are going wrong in the upper layers of the star. I've zoomed in to the region where things look the worst.
image
The biggest thing is the constant-ish few percent offset between the two continua.  The line cores match more closely (except in the IR), which indicates that the methods agree more in the upper layers, and disagree more near the roseland photosphere.  It's could that the difference comes down to Korg's vs MARCS's opacities (the default method "indexes" on the MARCS opacity at 500nm), but it is bigger than I would expect.  We should figure out how to test which is closer to the truth.

If the work exclusively with rectified spectra, the differences are much smaller (this was also shown in Wheeler+ 2023). Ideally, we would crank them down by another factor of 10 (or have a mode that does this), but it's acceptable for now.  
image

The only aspect of this that's really new is that I've now shown that it's the $\tau$ calculation that really matters.  The difference between integrating for $I$ using the linear scheme or the fancy bezier one is as small as we will likely every need.  That's 1) good for efficiency because we can use the efficient scheme with no guilt, and 2) fascinating, because it's exactly the opposite of the emphasis in [de la Cruz and Piskunov 2013](https://ui.adsabs.harvard.edu/abs/2013ApJ...764...33D) (originally proposed in Auer 2003) , which is where we get the bezier scheme from.  It must be that polarized 3D NLTE simulations, the opacity varies less linearly along rays, but I'm not sure which of those descriptors are the relevant ones.

We can also do a kind of crude convergence test by cutting out every-other atmospheric layer.  Unsurprisingly the anchored $\tau$ calculation is more robust here (though it's kind of cheating, since the MARCS calculation was done at full-res).  What's somewhat surprising is that the linear $I$ calculation appears slightly *more* robust than the bezier one.  I'll think about possible ways for this test to misfire.  It it holds up, this result is surprising but welcome.
image


ajwheeler:
# An update on performance

These times are for the radiative transfer calculation _only_, but the input absorption data came from these parameters:
```julia
linelist = [Korg.get_VALD_solar_linelist() ;  Korg.get_APOGEE_DR17_linelist()] 
atm = interpolate_marcs(5000.0, 3.0)
patm = Korg.PlanarAtmosphere(atm)
A_X = format_A_X()
wl_grid = [3000:0.01:5000, 16000:0.01:16500]
```

## old transfer implementation
This is for the default implementation (anchored + linear-flux-only in the new lingo).  I'm primarily concerned with the performance in the default case.
- planar (using $E_2$):  `282.709 ms (250013 allocations: 131.61 MiB)`
- spherical (using multiple rays)l: `1.541 s (4842296 allocations: 605.94 MiB)`

## new implementation
Again, I'm concerned with the defaults here, so I'm not evaluating bezier $\tau$ or $I$ integrals.
| | spherical | planar|
|-|-|-|
| **flux-only** | `1.778 s (6000348 allocations: 2.35 GiB)` |`292.917 ms (250040 allocations: 137.34 MiB)` |
| **$I$ retained**| `3.274 s (6000348 allocations: 4.81 GiB)` | `3.285 s (5000203 allocations: 4.66 GiB)` |

So there is a small slowdown for the default methods.  I'm resisting the urge to further optimize, though I will note that the increased allocations suggests to me that there's some low-hanging fruit here.

PR #260, opened 2024-02-07
Retry chemical equilibrium solver with very small initial $n_e$ if it fails to converge

ajwheeler:
I noticed while working on model atmosphere inversion that the Korg solver almost always converges if I start it with a very low $n_e$.  In this PR, I start the solver over with an initial guess of $n_e = n_\mathrm{tot} \times 10^{-10}$ if it failed the first time.  This change gets Korg to convergence for all parameters tried in #129.


ajwheeler:
Feels a bit silly to post these plots, bur for completeness:
image
image

PR #262, opened 2024-02-14
add Downgrade CI

ajwheeler:
Run test with lowest allowed versions of packages to try to make sure everything works.

https://www.stochasticlifestyle.com/semantic-versioning-semver-is-flawed-and-downgrade-ci-is-required-to-fix-it/


ajwheeler:
https://github.com/cjdoris/julia-downgrade-compat-action/issues/6#issuecomment-1950286456

PR #263, opened 2024-02-19
Precompute molecular cross section

ajwheeler:
This PR introduces functions to precompute molecular cross sections.  When applied to water lines in apogee, this decreases synthesis time by 5-100x.

Here's a plot of the error introduced by the procedure (for Teff=3800, logg=4.5). It's less than 4x10^-4 everywhere.  Technically this is an overestimate, as the precomputed cross-sections use more conservative line windows than the live synthesis.
image

TODO
- [x] handle vmic
- [x] functions for reading and writing cross-sections from hdf5 files
- [x] make sure everything works with noncontiguous wavelength ranges
- [ ] ~~handle wavelengths more gracefully?~~  Decided against this for now.


ajwheeler:
I don't know how to handle vmic sufficiently quickly via convolution at synthesis-time, so I've opted to add it as a dimension in the interpolator.  This requires more upfront CPU, and more memory, but it will still be a big improvement over doing all the molecular lines at synthesis time in certain cases.

image
Doing this testing revealed that (due to conservative line windows), this technique is actually slightly more accurate at larger vmics.  (I had to turn down `line_cutoff_alpha` for the comparison).

Issue #278, opened 2024-03-06
Something similar to MOOG weedout


ajwheeler:
Given a linelist, it would be very nice to be able to identify which lines contribute "at a given level", for some definition of that phrase.  There are at least two independent uses for this:
- pruning a linelist down to just the lines that you need.  This is done dynamically as synthesis proceeds, but doing it ahead of time will save lots of time for, e.g. molecular lines in hotter stars. (Requested by Thomas Nordlander)
- Annotating prominent features in a spectrum. (Requested by @MadiKenzie)

ajwheeler:
Closed by #281

PR #281, opened 2024-03-11
Weedout

ajwheeler:
This PR implements a function to identify all lines below a certain depth, were they synthesized in isolation.  It is an attempt to address #278.  

This implementation takes a first pass, identifying important lines by comparing their wavelength-integrated absorption coefficient to the total absorption, then does single-line synthesis on a more limited set of lines in the linelist.  This works well for line depths down to 1% (and possible shallower), but it's not very fast.  It's probably sufficient for now.

TODO
- [x] tests
- [x] documentation
- [x] deal with multiplets


ajwheeler:
OK, the finished version of this filters lines by looking at the ratio of the line-center alpha (estimated assuming a tophat profile with width equal to the Doppler width) to the continuum alpha.  This seems to work well.  Following that, the lines that make it through are ordered by EW.  This is so that you can run `prune_linelist` once and the choose exactly the threshold you want if you are making a plot.  If you don't order by EW, it saves time and the lines remain in wavelength order.

There is also a function `merge_close_lines` that merges line of the same species withing a configurable wavelength distance.  Rather than returning another vector of line objects (what would they even be?) it returns (wavelength, species) pairs, with the species as strings, not `Korg.Species` objects.  I don't love this interface, but I can't think of anything better.

Two general changes to the code made as part of this PR:
- handling of the wavelength arguments passed to `synthesize` has been split into a separte function, rather than having multiple methods of `synthesize`.   It will no loner fail if you pass a single wavelength range (not wrapped in a vector).
- `synthesize` has a new keyword argument, `use_chemical_equilibrium_from`, which will not recompute the chemical equilibirum if you pass a previously computed `sol`.  It's up to the user to ensure that the atmosphere and abundances are the same, and this is warned about in the docs.  I think this will be generally useful.

src/prune_linelist.jl
```diff
@@ -0,0 +1,141 @@
...
+
+Return the vector containing the strongest lines in  `linelist`, (optionally) sorted by approximate 
+equivalent width.
+
+# Arguments
+- `atm`: the atmosphere model
+- `linelist`: the linelist (a vector of `Line` objects)
+- `A_X`: the abundance of each element (see [`format_A_X`](@ref))
+- `wls...`: the wavelength ranges to synthesize over.  These are specified in the same for as the 
+  `wls` for [`synthesize`](@ref).
```
ajwheeler, line 14:
```suggestion
- `wls...`: the wavelength ranges to synthesize over.  These are specified the same way as the 
  `wls` for [`synthesize`](@ref).
```

src/prune_linelist.jl
```diff
@@ -0,0 +1,141 @@
...
+
+# Keyword Arguments
+- `threshold=0.1`: The threshold ratio in the line center absorption to the continuum absorption 
+   computed at the photosphere for a line to be included in the returned list.
+   `0.1` is a reasonable default for getting a sense of what might be measurable in a high-res, 
+   high-quality spectrum, but it should not be used to create a linelist for synthesis.  
+- `sort=true`: If `true`, the returned linelist will be sorted by approximate equivalent width.  If
```
ajwheeler, line 21: rename this kwarg

src/prune_linelist.jl
```diff
@@ -0,0 +1,141 @@
...
+        end
+        phot_ind = photosphere_indices[λ_ind]
+
+        E_upper = line.E_lower + c_cgs * hplanck_eV / line.wl 
+        levels_factor = exp(-β[phot_ind]*line.E_lower) - exp(-β[phot_ind]*E_upper)
+        # line center amplitude if it were a 1 Å tophat
+
+        σ = doppler_widths[line.species][phot_ind]
+        n_Z = n_div_Z[line.species][phot_ind]
+        α_λ_line_center = 1e8 * 10.0^line.log_gf*sigma_line(line.wl)*levels_factor*n_Z / σ
```
ajwheeler, line 79: factor of 1e8 here cancels with the one on line 55

src/prune_linelist.jl
```diff
@@ -0,0 +1,141 @@
...
+
+        if α_λ_line_center > threshold * cntm_sol.alpha[phot_ind, λ_ind]
+            push!(strong_lines, line)
+        end
+    end
+
+    # sort lines by approximate EW or leave them in wavelength order
+    if sort
+        approx_EWs = @showprogress "measuring $(length(strong_lines)) lines" map(strong_lines) do line
```
ajwheeler, line 88: make this turn-off-able

src/prune_linelist.jl
```diff
@@ -0,0 +1,141 @@
...
+        strong_lines[sortperm(approx_EWs, rev=true)]
+    else
+        strong_lines
+    end
+end
+
+"""
+    merge_close_lines(linelist; merge_distance=0.2)
+
+Turn a linelist into a list of the species and wavelengths of the lines in `linelist`, merging lines
```
ajwheeler, line 103: reword

src/prune_linelist.jl
```diff
@@ -0,0 +1,141 @@
...
+    else
+        strong_lines
+    end
+end
+
+"""
+    merge_close_lines(linelist; merge_distance=0.2)
+
+Turn a linelist into a list of the species and wavelengths of the lines in `linelist`, merging lines 
+of the same species that are within `merge_distance` (default: 0.2 Å).  This is useful for labelling
```
ajwheeler, line 104: spelling

src/prune_linelist.jl
ajwheeler: nm, not worth addressing

src/prune_linelist.jl
ajwheeler: not doing this because it conflicts with a synthesize kwarg

PR #296, opened 2024-05-03
Clarify carbon/alphas situation

ajwheeler:
Reword the docstring for `get_metals_h` to be less confusing.

Thanks @svenbuder for pointing this out.  I won't merge this PR until I hear back about whether this behavior is a problem for you.

### DIFF
--- a/src/synthesize.jl
+++ b/src/synthesize.jl
@@ -371,7 +371,7 @@ See also [`get_alpha_H`](@ref).
   convenience.
 - `ignore_alpha` (default: `true`): Whether or not to ignore the alpha elements when calculating 
   [metals/H].  If `true`, [metals/H] is calculated using all elements heavier than He.  If `false`, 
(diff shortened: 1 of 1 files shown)


svenbuder:
Hi Adam,

Having thought about it a bit more, I was wondering if maybe the initial value of carbon here might be the reason for the mismatch.

If carbon is not given as an argument, do you want it the c_h value to default to the fe_h or alpha_h value? I would assume the fe_h one right. Could you maybe double check this?

Thanks,


Sven

ajwheeler:
I think I didn't read your initial email carefully enough.

```julia
using Korg
fe_h = -0.5
alpha_h = -0.3
A_X = format_A_X(fe_h, alpha_h; solar_abundances=Korg.grevesse_2007_solar_abundances)
A_X[6] 
```
yields `7.890000000000001` on `v0.33.0` (which was only barely out when you emailed) and on this branch.  

On `v0.32.0` it's `8.09`, as you reported.

So I think `v0.33` actually solves this?

svenbuder:
Perfect! Thanks, Adam.

PR #301, opened 2024-06-03
Qfactor calculation contribution

andrew-saydjari:
This PR is tracking the contribution of methods of computing theoretical lower bounds on RV precision (different levels of approximation of the so-called Q-factor from Bouchy 2001). 

Happy to have feedback on code style etc. I plan to improve the function documentation. 

Currently, I am unhappy that the last two tests are not equal to eachother, because I was trying to compute the lower bound using the more rigorous function under the conditions that would make the approximation used in the second test true. I think the issue lies in the definition of the continuum and how it is interplaying with the S/N. The difference is only 2 m/s... but I will still run down.

### DIFF
--- a/src/qfactors.jl
+++ b/src/qfactors.jl
@@ -28,12 +28,12 @@ Compute the Q factor from the high resolution theoretical spectrum, the high res
 - `spec_hres::Vector{Real}`: High resolution theoretical spectrum
 - `wave_hres`: High resolution wavelength grid
 - `wave_lres`: Low resolution wavelength grid
-- `LSF_mat::Matrix{Real}`: LSF matrix
+- `LSF_mat`: LSF matrix
 
 # Keyword Arguments
 - `msk_lres::Vector{Bool}=nothing`: Mask for the low resolution spectrum to account for pixels masked from the observation
 """
-function Qfactor(spec_hres::Vector{Real},wave_hres,wave_lres,LSF_mat::Matrix{Real}; msk_lres::Vector{Bool}=nothing)
+function Qfactor(spec_hres,wave_hres,wave_lres,LSF_mat; msk_lres=nothing)
     nvecLSF = dropdims(sum(LSF_mat,dims=2),dims=2) # normalisation for the LSF
     spec_lres = LSF_mat*spec_hres./nvecLSF
 
@@ -58,14 +58,14 @@ Compute the RV precision using the true observed uncertainties on the continuum 
 - `spec_hres::Vector{Real}`: High resolution theoretical spectrum
 - `wave_hres`: High resolution wavelength grid
 - `wave_lres`: Low resolution wavelength grid
-- `LSF_mat::Matrix{Real}`: LSF matrix
+- `LSF_mat`: LSF matrix
 - `noise_contNorm::Vector{Real}`: Noise in the continuum normalized spectrum
 
 # Keyword Arguments
-- `msk_lres::Vector{Bool}=nothing`: Mask for the low resolution spectrum to account for pixels masked from the observation
+- `msk_lres=nothing`: Mask for the low resolution spectrum to account for pixels masked from the observation
 - `c::Real=299792458`: Speed of light in m/s
 """
-function precRV_MeasuredNoise(spec_hres::Vector{Real},wave_hres,wave_lres,LSF_mat::Matrix{Real},noise_contNorm::Vector{Real};msk_lres::Vector{Bool}=nothing,c::Real = 299792458)
+function precRV_MeasuredNoise(spec_hres,wave_hres,wave_lres,LSF_mat,noise_contNorm; msk_lres=nothing,c::Real = 299792458)
     nvecLSF = dropdims(sum(LSF_mat,dims=2),dims=2)
 
     dspec_dlam = zeros(length(spec_hres))
--- a/src/qfactors.jl
+++ b/src/qfactors.jl
@@ -1,0 +1,37 @@
+## Computation of Q-factors to provide an estimate of the RV precision for a given spectrum as a function of SNR
+
+## Making the approximation that the spectrum as an ``average'' SNR per pixel when estimating the RV precision
+function precRV_fromQ(Q,SNR,Npix;c = 299792458) # c in m/s, SNR per pixel
+    return c/(Q*sqrt(Npix)*SNR)
+end
+
+## Qfactor from the theoreticsl spectrum and LSF model, assuming uniform efficiency as a function of wavelength
+function Qfactor(spec_hres,wave_hres,wave_lres,LSF_mat; msk_lres=nothing)
+    nvecLSF = dropdims(sum(LSF_mat,dims=2),dims=2) # normalisation for the LSF
+    spec_lres = LSF_mat*spec_hres./nvecLSF
+
+    dspec_dlam = zeros(length(spec_hres))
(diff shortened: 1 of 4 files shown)


src/qfactors.jl
```diff
@@ -0,0 +1,80 @@
...
+
+Compute the RV precision from the Q factor, the SNR per pixel, and the number of pixels in the spectrum.
+
+# Arguments
+- `Q::Real`: Q factor of the spectrum
+- `SNR::Real`: SNR per pixel of the spectrum
+- `Npix::Real`: Number of pixels in the spectrum
+
+# Keyword Arguments
+- `c::Real=299792458`: Speed of light in m/s
```
ajwheeler, line 15: You can use `Korg.c_cgs` (unqualified) for this.

src/qfactors.jl
```diff
@@ -0,0 +1,80 @@
...
+## Qfactor from the theoreticsl spectrum and LSF model, assuming uniform efficiency as a function of wavelength
+"""
+    Qfactor(spec_hres,wave_hres,wave_lres,LSF_mat; msk_lres=nothing)
+
+Compute the Q factor from the high resolution theoretical spectrum, the high resolution wavelength grid, the low resolution (observed) wavelength grid, and the LSF matrix.
+
+# Arguments
+- `spec_hres::Vector{Real}`: High resolution theoretical spectrum
+- `wave_hres`: High resolution wavelength grid
+- `wave_lres`: Low resolution wavelength grid
```
ajwheeler, line 30: Style-wise, can you rename these `synth_flux`, `synth_wl`, and `obs_wl`?

src/qfactors.jl
```diff
@@ -0,0 +1,80 @@
...
+- `LSF_mat`: LSF matrix
+- `noise_contNorm::Vector{Real}`: Noise in the continuum normalized spectrum
+
+# Keyword Arguments
+- `msk_lres=nothing`: Mask for the low resolution spectrum to account for pixels masked from the observation
+- `c::Real=299792458`: Speed of light in m/s
```
ajwheeler, line 66: ditto

src/qfactors.jl
```diff
@@ -0,0 +1,80 @@
...
+    else
+        return sqrt(sum(Wvec[msk_lres])/sum(spec_lres[msk_lres]))
+    end
+end
+
+## Theoretical limit on RV precision from the measured noise in the spectrum
+"""
+    precRV_MeasuredNoise(spec_hres,wave_hres,wave_lres,LSF_mat,noise_contNorm;msk_lres=nothing,c = 299792458)
+
+Compute the RV precision using the true observed uncertainties on the continuum normalized spectrum as a function of wavelength.
```
ajwheeler, line 55: Can you add a link to the Q factor paper?

src/qfactors.jl
```diff
@@ -0,0 +1,80 @@
...
+- `spec_hres::Vector{Real}`: High resolution theoretical spectrum
+- `wave_hres`: High resolution wavelength grid
+- `wave_lres`: Low resolution wavelength grid
+- `LSF_mat`: LSF matrix
+- `noise_contNorm::Vector{Real}`: Noise in the continuum normalized spectrum
+
+# Keyword Arguments
+- `msk_lres=nothing`: Mask for the low resolution spectrum to account for pixels masked from the observation
+- `c::Real=299792458`: Speed of light in m/s
+"""
```
ajwheeler, line 67: Can you throw a `See also: ` section at the end of these three docstrings linking to the other two functions?

test/qfactors.jl
```diff
@@ -0,0 +1,30 @@
+@testset "qfactors" begin
+    # Check a Q factor gives the right RV precision
+    @test abs(Korg.precRV_fromQ(1100,100,8000)-30.470741601352298) < 1e-10
```
ajwheeler, line 3:
```suggestion
    @test Korg.precRV_fromQ(1100,100,8000) ≈ 30.470741601352298 atol=1e-10
```

ajwheeler: This is being extraordinarily picky, sorry

test/qfactors.jl
```diff
@@ -0,0 +1,30 @@
...
+
+    # Check that a solar spectrum at APOGEE resolution gives the expected Q factor
+    delLog = 6e-6; 
+    wavetarg = 10 .^range((start=4.179-125*delLog),step=delLog,length=8575+125)
+    minw, maxw = extrema(wavetarg);
+    
+    wl_lo, wl_hi, wl_step = 15_000, 17_000, 1//100
```
ajwheeler, line 10: Is there a particular reason to use a `Rational` here?

src/qfactors.jl
```diff
@@ -0,0 +1,80 @@
...
+function precRV_MeasuredNoise(spec_hres,wave_hres,wave_lres,LSF_mat,noise_contNorm; msk_lres=nothing,c::Real = 299792458)
```
ajwheeler, line 68: I would probably name this something wordier like `RV_prec_from_noise`, and similarly `RV_prec_from_Q`.  You don't have to use my names, but I feel like the combination of snake and camel case is awkward.

test/qfactors.jl
```diff
@@ -0,0 +1,30 @@
...
+    tspec = (kout.flux)./(kout.cntm);
+    Q = Korg.Qfactor(tspec, x_model, wavetarg, LSF_model)
+    @test abs(Q-1182.420317367743)<1e-10
+
+    msk = ones(Bool,length(wavetarg))
+    msk[1:100].=false
+    msk[end-100:100].=true
+    @test abs(Korg.precRV_MeasuredNoise(tspec, x_model, wavetarg, LSF_model, 1 ./100*ones(length(wavetarg)),msk_lres=msk)-29.490659698846365)<1e-10
+    @test abs(Korg.precRV_fromQ(Q,100,count(msk)).-27.34006821159959)<1e-10
+    # I think these two should agree... but the problem must be from the continuum defintion and how that propagates to S/N
```
ajwheeler, line 29: I want to understand this before we merge if possible.

andrew-saydjari: Yes. Same. High on my to do.

test/qfactors.jl
andrew-saydjari: Yes. Because for chi2 surface searching and interpolation, float precision issues are realy really annoying and have caused problem for apMADGICS before. So I want to make sure you are explicitly testing with rationals more than you are. I am taking this moment and my liberty in these tests to incept that.

src/qfactors.jl
andrew-saydjari: I will stop passing this as a keyword then and do Korg.c_cgs*1e-2

src/qfactors.jl
andrew-saydjari: it is... I am terrible about that.

src/qfactors.jl
andrew-saydjari: done!

src/qfactors.jl
andrew-saydjari: I decided to put that on the Q factor docstring and because the see alsos link there, I am fine with that.

ajwheeler:
I don't think I can push commits to this without getting permission to push to your fork.  (If someone knows better let me know.) I'll open a new PR.

Issue #300, opened 2024-05-28
Highlighting high memory usage for `fit_spectrum`


shbhuk:
Running the Fitting Spectrum example based on Griffith 2022 utilizing `fit_spectrum` seems to require a lot of RAM and gives an OutofMemoryError(). Based on inputs from Adam Wheeler and checking the documentation for `fit_spectrum`, it seems like the culprit might be my Julia version --- `1.9.2`. Will update it to `1.10` and re-try.

Example - https://github.com/ajwheeler/Korg.jl/blob/main/misc/Tutorial%20notebooks/fitting/Fit%20spectrum.ipynb
Documentation - https://ajwheeler.github.io/Korg.jl/stable/API/#Korg.Fit.fit_spectrum

ajwheeler:
Thank you for posting this for others to see.  I'll add that in many cases, you won't get an `OutOfMemoryError`, and may instead see that the code runs much more slowly than it should according to the tutorial.

I'll wait for you to confirm that upgrading to Julia 1.10 solves this before closing.

shbhuk:
I upgraded to Julia 1.10 and reran the example. Firstly, truncating the spectrum to < 5000 A and then comparing the before and after, the memory usage is similar but time taken is halved.

When running the original example (with the full spectra) it runs fine once, but when trying to re-run the exact same cell, gives me the same `OutofMemoryError` while processing the SparseArrays while computing the LSF matrix. 

I was running things locally before shifting to a Linux HPC but it seems like there might be some memory leak somewhere? 

![image](https://github.com/ajwheeler/Korg.jl/assets/25436514/471f4a2a-8832-49a7-894d-2f4c5da37f5f)

Old Julia version: 1.9.2
New Julia version: 1.10.3
Korg version 0.33.0
RAM 32 GB

shbhuk:
Sorry, I should clarify. I haven't yet shifted to an HPC. The plan was to test things locally before batch processing on the HPC. So this could be a Windows peculiarity? 

Project Korg v0.33.0
Status `C:\Users\skanodia\Documents\GitHub\Korg.jl\Project.toml`
  [336ed68f] CSV v0.10.14
  [717857b8] DSP v0.7.9
  [a93c6f00] DataFrames v1.6.1
  [163ba53b] DiffResults v1.1.0
  [442a2c76] FastGaussQuadrature v1.0.2
  [f6369f11] ForwardDiff v0.10.36
[...]

PR #302, opened 2024-06-04
profile memory usage in test suite

ajwheeler:
See #300. There appears to be a memory usage problem on windows only.  Until I can get my hands on a windows machine, I'm going to abuse github actions with the PR to try to figure out what's going on.


ajwheeler:
I'm renaming this PR because it's not reproducing the problem and I plan to merge this into main because I like outputting the info.  For posterity (because the test suite output will eventually go away), I'll note that the allocations in the test suite are the same on windows, macOS, and linux.  I think this means that it's probably a problem with the garbage collection?

PR #304, opened 2024-06-11
Qfactor

ajwheeler:
continuation of #301


ajwheeler:
Btw, sorry for completely rewriting the test.  I didn't mean to but it slowly drifted away from the original while I was debugging it.

test/qfactors.jl
```diff
@@ -0,0 +1,50 @@
...
+    atm = Korg.interpolate_marcs(5777, 4.4, A_X)
+    sol = synthesize(atm, apolines, A_X, synth_wls);
+
+    synth_flux = (sol.flux)./(sol.cntm);
+    # this denominator is a normalization factor
+    flux = (LSF_model * synth_flux) 
+
+    # for the Q factor calculation to be exact, the noise must be photon-dominated
+    # (This is approximating the continuum as flat)
+    obs_err = sqrt.(flux) .* 0.0456
```
andrew-saydjari, line 32: why is this here, it is overwritten by L36 right?

andrew-saydjari: As long as we can delete this line, then I am very happy. I don't think the test looks rewritten at all. It is exactly in the spirit I had.

ajwheeler, line 32:
```suggestion
```

ajwheeler: good catch!

ajwheeler:
I think github actions might be haunted.  If this doesn't pass, I'm going to investigate a bit before merging.

PR #305, opened 2024-06-17
Gaia-ESO survey linelist

ajwheeler:
This adds another built-in linelist.  It's distributed via the artifacts system, but downloaded lazily, because it's kinda big.  It also slightly speeds-up the winnowing of the linelist passed to `synthesize` to include only lines in the right wavelength ranges.  (This could definitely be improved further.)


data/linelists/GES/repackage_GES_linelist.jl
```diff
@@ -0,0 +1,72 @@
+using Korg, HDF5, FITSIO
+
+# this should point to the base dir of the Korg_data repo
+# https://github.com/ajwheeler/Korg_data
+Korg_data_dir = "../../../../Korg_data"
```
mabruzzo, line 5: Feel free to ignore this!
```suggestion
Korg_data_dir = joinpath(@__DIR__,"../../../../Korg_data")
```

mabruzzo (approved):
I had 1 minor suggestion (feel free to ignore it!), but this looks great to me!

data/linelists/GES/repackage_GES_linelist.jl
ajwheeler: Thanks!

PR #306, opened 2024-06-18
normalize LSF matrix in a more memory-efficient way

ajwheeler:
Hopefully solve #300 (details tk)

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -173,7 +173,14 @@ function compute_LSF_matrix(synth_wls::AbstractVector{<:Real}, obs_wls, R; windo
         @. convM[i, r] += ϕ
         next!(p)
     end
-    convM .* normalization_factor
+    if renormalize_edge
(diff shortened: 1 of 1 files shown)


ajwheeler:
This change fixes the memory issue documented in #300, which arises only on windows.  I was able to recreate the behavior and have confirmed that this fixed it, at least on the machine I an using.

It turns out that the problem is not actually anything to do with fitting specifically, merely the final step in constructing the LSF matrix, normalization.  I was using a broadcasted multiplication which turned out to be very memory inefficient because of the inner workings of sparse arrays.  If you look at the diff for this PR you'll see that the change is very small.  It results in in many many fewer allocations.

I'm not entirely certain why this memory problem was triggered only on windows, but I think it must have something to do with the behavior of the garbage collector / memory allocation system.  Benchmarking the code on any system shows the same total memory allocated, but the instantaneous footprint of the process doesn't grow at all on macOS or linux.  On windows it simply fails.

Issue #309, opened 2024-07-01
GES linelist type inference crash


ajwheeler:
When working on #305, I got a weird error (`fatal error in type inference (type bound)`).  After some cursory googling, I "fixed" this by explicitly (and redundantly) converting several parameter to Float64s.  Unfortunately the problem has reappeared https://github.com/ajwheeler/Korg.jl/actions/runs/9717153115/job/26822219107

I have not been able to reproduce this outside of github actions on my computer, including with x64 julia running via Rosetta.

ajwheeler:
Seems to be fixed.

Issue #310, opened 2024-07-01
 $\alpha_{5000}$ and line opacity in M stars


ajwheeler:
In Korg, by default, the formal solution (radiative transfer) uses $\tau_{5000}$, the optical depth at 5000 Å.  In order to do this it must calculate the absorption coefficient at the same wavelength, $\alpha_{5000}$.  

At present, it ignores contributions to $\alpha_{5000}$ from lines.  This is an unproblematic choice, except for M stars, where you get line blanketing across the visible.

I think the right way to handle this is to use the line opacity when it's available (when the linelist encompasses 5000 Å), and to not worry about it (for now) when it isn't.  Plots to come.

ajwheeler:
As with many detail of how RT is done (see #257), this impacts the continuum more than the rectified flux. (It's defintiely still important for both though. (And, as a side note, it's interesting to see that RT choices in general are crucial for cool stars, easily at the multi percent level.)

![image](https://github.com/ajwheeler/Korg.jl/assets/711963/ecfbc219-3be1-4468-a746-035d40a7106a)
![image](https://github.com/ajwheeler/Korg.jl/assets/711963/714c6032-4619-4494-a6c6-4d0b1770b9ea)

At slightly warmer Teff, things are less dicey, but still a but dicey:
![image](https://github.com/ajwheeler/Korg.jl/assets/711963/dd770121-2d79-4340-a28d-fdfa5e47f489)
![image](https://github.com/ajwheeler/Korg.jl/assets/711963/f08b8147-e8bf-43db-8d97-fd7a2c1937d8)

ajwheeler:
Here's what this looks like in the IR:
![image](https://github.com/user-attachments/assets/620a09b4-1999-4137-98ea-a02df2be18d0)
![image](https://github.com/user-attachments/assets/9b9075f4-7761-459e-8e89-343ff7c93743)

ajwheeler:
Closed by #313

PR #311, opened 2024-07-02
skip failing GES test for now

ajwheeler:
Skipping the GES test for now because I don't like having a failing test badge in the github repo.  #309 will remain open, obviously.

Open to not doing this if people think it's inexcusable vanity, etc.

### DIFF
--- a/src/linelist.jl
+++ b/src/linelist.jl
@@ -616,6 +616,7 @@ The Gaia-ESO survey linelist from
 [Heiter et al. 2021](https://ui.adsabs.harvard.edu/abs/2021A&A...645A.106H/abstract).
 """
 function get_GES_linelist()
+    @warn "This function is may fail on some systems. See https://github.com/ajwheeler/Korg.jl/issues/309 for details."
     path = joinpath(artifact"Heiter_2021_GES_linelist", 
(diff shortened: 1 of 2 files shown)


mabruzzo:
I understand the impulse. As I understand it, this is only an issue with the Gaia ESO linelist parser introduced by PR #305, right? If so, then I think that this is fine. But, I think I would have the Gaia ESO logic raise some kind of warning saying that ``@warn "This function is experimental and not currently tested"`` or something.

andycasey:
I dunno, spectroscopists are particularly good at ignoring warnings when they shouldn’t.

mabruzzo:
I don't know if it's just spectroscopics. I'm also very good at ignoring warnings. I love typing ``python setup.py develop`` and I've gotten warnings for years!

I think a warning here is better than nothing (it's quick and easy). If we are really that worried about people ignoring warnings, then we probably should be renaming the function (or relevant kwarg) to make it clear that it is experimental OR we should remove it altogether

ajwheeler:
In case, I'm not worried about people ignoring the warning, because the bug causes the function to error out rather than returning something wrong.  I hope to have this fixed within the next week or so (optimistic), but I want to prioritize #310 and working on a couple drafts.

PR #313, opened 2024-07-03
include line contributions to `alpha_5000`

ajwheeler:
This addresses #310, and was used to generate the plots in that thread.

Running the linelist filter an additional time to get the lines around 5000 \AA is not free, and we may want to consider doing this smarter.  It takes the run time from 7.5s to 8s when synthesizing the Teff=3500, logg=4.5 spectra.


ajwheeler:
The most recent commit allows for pretabulated molecular cross sections to be for wavelengths which don't exactly match those used for synthesis. (This makes it much nicer to apply them to `alpha_5000`.) Linear interp is used to put them on the synthesis wavelength grid, and they are not extrapolated outside the region in which they were calculated.

ajwheeler:
This implementation is (I think) finished.  It will work correctly regardless of what the user does. In other words, there are no gotcha cases that require knowledge of this issue.  It adds as estimated 0.05s to synthesis.

ajwheeler:
Reviews welcome, but since this is fairly pressing, I will reread the diff and merge myself early next week.

PR #314, opened 2024-07-09
add parameter to keep lines outside of wl range in prune_linelist

ajwheeler:
For specific use cases this is helpful.


ajwheeler:
Merged into #313

PR #316, opened 2024-07-12
use correct indices when normalizing LSF matrix

ajwheeler:


ajwheeler:
This fixes a bug introduced by #306.

PR #317, opened 2024-07-24
allow user to specify indidual mu values

### LINKED ISSUE #191: Calculating intensity at user-specified angle
Feature request from @RozanskiT: calculating $I$ at the top of the atmosphere at arbitrary $\mu$.

The current situation is that users can specify the number of $\mu$ points to use, but the specific values are chosen by a quadrature scheme.  Along with the flux, the intensity at each of these angles is returned by `synthesize`.

I think the best way to address this is to make it possible for users to specify *either* the number of $\mu$ values to use, *or* the specific values themselves.  In the latter case, flux can be obtained via trapezoidal integration, and the documentation would be clear that it is on the user to make good choices if they care about the flux being accurate.

ajwheeler:
closes #191

### DIFF
--- a/src/RadiativeTransfer/RadiativeTransfer.jl
+++ b/src/RadiativeTransfer/RadiativeTransfer.jl
@@ -6,7 +6,7 @@ using FastGaussQuadrature: gausslegendre
     generate_mu_grid(n_points)
     generate_mu_grid(μ_values)
 
-Used by both radiative transfer schemes to compute quadature over μ. Returns `(μ_grid, μ_weights)`.
+Used by both radiative transfer schemes to compute quadrature over μ. Returns `(μ_grid, μ_weights)`.
 If an integer is passed, generate a grid of Gauss-Legendre quadrature points of corresponding size.
 Otherwise, use the provided μ values.
 """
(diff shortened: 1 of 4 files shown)


src/RadiativeTransfer/RadiativeTransfer.jl
```diff
@@ -1,19 +1,29 @@
...
+Otherwise, use the provided μ values.
 """
-function generate_mu_grid(n_points)
+function generate_mu_grid(n_points::Integer)
     μ_grid, μ_weights = gausslegendre(n_points)
     μ_grid = @. μ_grid/2 + 0.5
     μ_weights ./= 2
     μ_grid, μ_weights
 end
+function generate_mu_grid(μ_grid::AbstractVector{<:Real})
```
mabruzzo, line 19: Out of curiosity, is a "range" handled by this case? (it probably doesn't matter)

mabruzzo (approved):
LGTM (I had 1 minor question -- mostly from curiosity). Feel free to merge whenever!

src/RadiativeTransfer/RadiativeTransfer.jl
ajwheeler: Yes!

mabruzzo: Neat!

PR #323, opened 2024-08-12
[WIP] Behaviour with `Korg.Fit.ews_to_stellar_parameters` when no EW errors given 

andycasey:
The documentation regarding `measured_EWs` for `Korg.Fit.ews_to_stellar_parameters` says that if no EW errors are specified, all lines are assumed to have the same uncertainty. This is kinda true but leads to some unusual behaviour.

All lines are assumed to have the same uncertainty in EW of 1 mA (`measured_EW_err=ones(length(measured_EWs))`).

This sets equivalent uncertainties in EW, which is not the same as equivalent uncertainties in abundance, because strong lines (e.g., 200 +/- mA) end up getting weighted more than weak lines (e.g., 20 +/- 1 mA). 

This is a WIP for now because I think Adam would prefer a `default_EW_error` function to set the default value.

### DIFF
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -562,7 +562,7 @@ A tuple containing:
 - `max_iterations` (default: 30) is the maximum number of iterations to allow before stopping the 
    optimization.
 """
-function ews_to_stellar_parameters(linelist, measured_EWs, measured_EW_err=ones(length(measured_EWs)); 
+function ews_to_stellar_parameters(linelist, measured_EWs, measured_EW_err=nothing; 
                                    Teff0=5000.0, logg0=3.5, vmic0=1.0, m_H0=0.0,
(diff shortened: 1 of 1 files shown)


ajwheeler:
Do you still believe in this given what we now know about EW fitting?

andycasey:
I do because the estimated error on EW is usually fractional (a percentage), but I'd like to do some science tests with the revised EW methods before strictly proposing how EWs are handled when no uncertainties are given.

If you are keen to do housekeeping on GitHub issues then you can close this PR and add it to an item in the roadmap/TODO list.

ajwheeler:
Happy to leave it open as long as we still want to investigate.

ajwheeler:
We should revisit this now that #331 is ready-ish.

Is there a science case/test that you think I should look into to be convinced? One possibility might be changing the default to something larger than 1 mA?

Issue #324, opened 2024-08-15
Issue with get_metals_h ? 


segasai:
Hi, 
 
I'm trying to understand this part of the code:

https://github.com/ajwheeler/Korg.jl/blob/9e6db774ea2497e347364cbb854946b11374371d/src/synthesize.jl#L453

and I cannot see why this is correct, given the alpha elements are 8:2:22
Shouldn't it be  [3:6 ; 7:2:21 ; 23:MAX_ATOMIC_NUMBER]  or am I missing something ? 

Thanks

ajwheeler:
You are correct.  I am going to fix this and refactor/write tests to try to eliminate this class of bug.

ajwheeler:
#327 fixes this, and will make similar bugs harder to introduce in the future.

PR #327, opened 2024-09-05
refactor to ensure consistent definition of alpha elements

ajwheeler:
This fixes the bug pointed out in #324.


ajwheeler:
The first commit is just breaking abundance stuff into a separate file. Excluding it when viewing this diff makes the real changes clear.

mabruzzo (approved):
Looks great to me!

PR #329, opened 2024-09-13
Slurm example

### LINKED ISSUE #303: documentation for generating grids on a SLURM cluster
@yingyisong

ajwheeler:
Add documentation showing how to generate a grid on a slurm cluster.  Closes #303.

# TODO
- [x] add to docs
- [x] plots spectra and make sure they are sane


ajwheeler:
There are still a few things to decide/figure out.  @andrew-saydjari, any input is welcome.
- This is currently manually calling the garbage collector after each synthesis.  This isn't something I normally do, so I should probably test its effects.
- `SlurmClusterManager.jl` seems to be neglected, perhaps I should convert this to use `ClusterManagers.jl`
- I'm not sure why the last line, `rmprocs(workers())` is needed

andrew-saydjari:
- I was having some memory leak problems a LONG time ago, and saw no performance change in terms of synthesis speed for adding the line, with better RAM usage stability. If you can determine that it is no longer needed, happy to remove.
- How is it being neglected? addprocs(SlurmManager()) is a direct call to SlurmClusterManager.jl
- It can be helpful for job completion to remove all of the worker processes so that slurm does not have to kill them. It just makes the exit a bit cleaner.

ajwheeler:
I'm leaving in the `rmprocs(workers())`. Thanks for the tip.

I tried generating a few thousand spectra (enough to saturate a node for ~ 5 minutes) with and without the manual GC call and got extremely similar peak memory usage, so I'm going to remove the GC call here just for the sake of keeping things simple, but I'll keep an eye on it as I run things.

PR #331, opened 2024-09-17
Saturated line EWs

ajwheeler:


ajwheeler:
In the spirit of transparency, work on this PR is indicating that the EW precision result in the second Korg paper is probably flawed.  If you are using Korg for EWs, I suggest holding off.  I plan to have this resolved, communicated better, etc very soon.

ajwheeler:
The working version of this doesn't do any fancy tricks (the kind discussed in Gray and used by MOOG) involving fictitious lines.  I messed around with that for a while and could never get it to work.

In the end the solution which was fastest and also most accurate (in the sense that the inferred per-line abundances could be plugged back in to correctly recover the equivalent widths) is a simple custom solver that takes newton steps towards the best-fit abundances for each line.  The reason it's custom is that it enables me to do the syntheses simultaneously, while keep the optimization of each line mathematically separate.  This ends up being significantly faster than using a fancier/better nonlinear optimization algorithm serially on each line.

The algorithm for stellar parameter estimation is largely untouched, but it now works in two phases: the initial fast phase uses the old ews->abundances algorithm (the one that assumes all lines are on the linear part of the curve of growth).  This is fast, and gets you surprisingly close to the right answer.  After than, it refines with the new, correct ews->abundances.

Stellar parameter estimation could be sped up significantly.  Right now it involves solving a system of equations, wherein each residuals evaluation requires a nonlinear solve.  Two ways to make this better:
- instead of calculating the residuals with dual numbers, use floats, then use math to get the partials and reconstruct the duals of the optimum.  I use this trick in to speed things up when autodiffing through Korg's chemical equilibrium solver.
- Alternatively, reformulate the whole problem as a joint optimization of the per-line abundances and the stellar parameters.  I think this could done so that the solution is identical up to numerics, perhaps via the complete linearization of the problem.

Another shortcoming of this implementation is that it only goes down to [m/H] of -2.5, the bottom of the SDSS MARCS grid.  Using the low-metallicity grid instead isn't hard in principle, it's just finicky since they have difference logg boundaries.

src/Fit/fit_via_EWs.jl
```diff
@@ -0,0 +1,507 @@
...
+
+    if exact_calculation
+        A = ews_to_abundances(atm, linelist, A_X, EW; vmic=vmic, blend_warn_threshold=Inf,
+                              passed_kwargs...)
+    else
+        A = ews_to_abundances_approx(atm, linelist, A_X, EW; vmic=vmic, blend_warn_threshold=Inf,
+                                     passed_kwargs...)
+    end
+    # convert error in EW to inverse variance in A (assuming linear part of C.O.G.)
+    A_inv_var = (EW .* A ./ EW_err) .^ 2
```
ajwheeler, line 486: This is wrong.  But probably not so much so that it should block this PR.  I think it can be fairly safely dealt with afterwards.

src/line_absorption.jl
```diff
@@ -36,10 +36,10 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
     β = @. 1 / (kboltz_eV * temps)
 
     # precompute number density / partition function for each species in the linelist
-    n_div_Z = map(unique([l.species for l in linelist])) do spec
+    n_div_U = map(unique([l.species for l in linelist])) do spec
```
ajwheeler, line 39: This change is purely for terminological consistency.  Partition functions are spelled "`U`" in the Korg codebase

src/linelist.jl
```diff
@@ -89,6 +95,12 @@ struct Line{F1,F2,F3,F4,F5,F6}
                                                                         gamma_stark, vdW)
     end
 end
+# constructor to allow for copying a line and modifying some values (see docstring)
```
ajwheeler, line 98: Worth nothing that this PR introduces a new constructor for `Line`s.

PR #334, opened 2024-10-01
arbitrary postprocessing in fit_spectrum 

ajwheeler:
At the moment, this does one thing, which is to allow the caller to pass in a function to arbitrarily postprocess the synthesized spectrum.

(This is useful to apply a constant continuum shift, but note that there are also params for that which I should consider removing...)

### DIFF
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -228,7 +228,7 @@ values are used.
     achieved by the optimizer is about 10x bigger than this.
   - `postprocess` can be used to arbitrarilly transform the synthesized (and LSF-convolved) spectrum
     before calculating the chi2.  It should take the form `postprocess(flux, data, err)` and write
-    it's changes in-place to the flux array.
+    its changes in-place to the flux array.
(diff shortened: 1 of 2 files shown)


src/fit.jl
```diff
@@ -226,7 +226,10 @@ values are used.
     the default value, `1e-4`, provides a theoretical worst-case tolerance of about 0.15 K in `Teff`,
     0.0002 in `logg`, 0.0001 in `m_H`, and 0.0004 in detailed abundances. In practice the precision
     achieved by the optimizer is about 10x bigger than this.
-    Any additional keyword arguments will be passed to [`Korg.synthesize`](@ref) when synthesizing the
+  - `postprocess` can be used to arbitrarilly transform the synthesized (and LSF-convolved) spectrum
+    before calculating the chi2.  It should take the form `postprocess(flux, data, err)` and write
+    it's changes in-place to the flux array.
```
ajwheeler, line 231:
```suggestion
    its changes in-place to the flux array.
```

PR #336, opened 2024-10-02
Wavelengths

ajwheeler:
Significant refactor to try to consolidate all the ways the noncontinguous ranges of wavelengths get handled.  This is going to be annoying but it's a needed abstraction.

This introduces a type `Korg.Wavelengths`, which is an `AbstractArray` implemented under the hood as a vector of ranges.  This means that finding values can be done very quickly, but elements can also be read out like it's a flat array.  This eliminates the need to carry around `wls` and `wl_ranges` in many places.  It also centralizes a bunch logic around automatic array->range conversion, air to wavelengths conversion, etc.


ajwheeler:
Thanks!  I've been wanting something like this for a while and I finally got around to implementing it on Friday.  It's kinda hacky but it works.

ajwheeler:
I've also significantly sped up the computation of LSF matrices (so much so that I removed the progress bar), by assigning to the along the fast axis (cols) and then doing an lazy adjoint at the end to keep the shape the same.
image
This used to take about a minute.

ajwheeler:
I'm also removing `Korg.rectify`, which is an unnecessary footgun, and which was the solve caller of `move_bounds` after this refactor.

ajwheeler:
image

!

ajwheeler:
If anyone is inclined to review this by just looking at the changes to the `synthesize` and `Korg.Fit.fit_spectrum` docstrings, that would be very valuable.

PR #368, opened 2024-11-20
CUDA

ajwheeler:
Do the line absorption calculations on GPU.


ajwheeler:
I think for now I will just call this an engineering demo and reopen if/when we decide to pursue further.

Issue #365, opened 2024-11-18
float issues with wavelength axis in synthesize


andrew-saydjari:
Here is a reproducer

```
ex_atm = Korg.interpolate_marcs(5750, 2.0, 1.0)
ex_spec = synthesize(ex_atm, lines, format_A_X(1.0), 3805.0, 5645.0)
extrema(ex_spec.wavelengths)
```

The result is :
``
(3805.0000000000005, 5645.0)
``

andrew-saydjari:
```
ex_spec = synthesize(ex_atm, lines, format_A_X(1.0), 3805, 5645, 1)
extrema(ex_spec.wavelengths)
```

Also fails

ajwheeler:
I will keep this in mind as something to improve on, but tbh it doesn't seem critical.  It's already the case that to analyze real data you must buffer the wl range on either side in order to get the LSF right.

andrew-saydjari:
I just hate Float issues like this and try to avoid them because I find they cause no end of annoying edge cases that bother me later down the road (and can be hard to run down). It is also non-Julian to not respect my type choice if I pass Ints. 

Off my philosophical soapbox, I agree with your assessment of where your time investment is valuable and think you should back burner it, but keep the issue open.

ajwheeler:
Definitely not claiming it's Julian. I would like to fix this at some point. Just being honest that it's not a priority. I'll definitely leave the issue open.

ajwheeler:
Hopefully the range stuff the Stefan presented at juliacon will fix this.

Issue #356, opened 2024-11-06
too-strong CH lines in Gaia ESO survey linelist?


ajwheeler:
Sarah Aquilina recently brought to my attention that when she synthesized a small window using the GES linelist, her pseudocontinuum was offset from was she got synthesizing a large window.  This indicates the influence of distant lines, and turning up the `line_buffer` when synthesizing in the small window confirmed that.

I ran some checks with subsets of the GES linelist and determined that the culprit is a few (~tens) CH lines with weirdly large oscillator strengths.  I have checked that a couple of these are actually present in the Heiter et al tables, so I don't _think_ I'm making a parsing error.  
<details><summary> Here's a subset:</summary>
<img width="1179" alt="image" src="https://github.com/user-attachments/assets/ed681059-a7bc-40f0-b9b4-1f153d76c01f">
</details>

If I plot a histogram of $\log gf$ values of CH lines in the GES linelist table alongside the same histogram from Masseron 2014, which is the source.  It also looks like too-strong lines may have been inserted erroneously.  Note that the Masseron table include things outside the wavelength range of the GES linelist, and that other cuts may have been made, so we shouldn't expect an exact correspondence.
![image](https://github.com/user-attachments/assets/28bc93d0-efb8-4d19-9d67-718387fd5deb)

PR #370, opened 2024-11-20
filter out CH lines, because they appear to be problematic (see #365)

ajwheeler:
Addresses #356.  It would be better to fix the underlying problem, but removing the CH lines is 90% of a solution.


ajwheeler:
I've edited this to only remove the too-strong CH lines.  This is what's left, which looks like it is correct (teff = 5500, logg=4.5).

In order to get this (and not have a pseudocontinuum offset, I cut all CH lines with loggf > -1.9.

image

PR #373, opened 2024-11-20
faster filter_linelist

ajwheeler:
This is a much faster implementation, and it means that it is now much, much less important to pre-filter your linelist to the wavelength range.


mabruzzo:
> image
> 
> !!!

It's beautiful

ajwheeler:
This retains too many lines when there are multiple wl windows.

mabruzzo (commented):
What are the contexts where `searchsortedfirst` and `searchsortedlast` return `nothing`? The documentation says nothing about this. (And looking at the source code, there doesn't seem to be a branch that produces `nothing`)

ajwheeler:
You are right, I misremembered that.

ajwheeler:
The bug here is that lines can get included more than once via multiple windows (when buffered by `line_buffer`).

PR #375, opened 2024-11-21
Auto adjust continuum in each window

ajwheeler:
This PR introduces a kwarg to `fit_spectrum` which, at each step of the optimizer, shifts the model spectrum by the MLE linear adjustment to the observed continuum in each window.

TODO: add tests


ajwheeler:
This also fixes two subtle bugs in the handling of synthesis wls for windows fits.  These could cause small parameter biases (10s of K, metallicity differences of 0.1*a few)

Issue #378, opened 2024-12-01
domain error with radiative transfer


segasai:
Hi, 

I don't know if it's worth reporting, but I looked at korg failures over large grid that are not 'chemistry convergence' issues. 
And this 
 thing stand out: 

```
ERROR: LoadError: DomainError with -7.986061283680538e-6:
log was called with a negative real argument but will only return a complex result if called with a complex argument. Try log(Complex(x)).
Stacktrace:
  [1] throw_complex_domainerror(f::Symbol, x::Float64)
    @ Base.Math ./math.jl:33
  [2] _log(x::Float64, base::Val{:ℯ}, func::Symbol)
    @ Base.Math ./special/log.jl:301
  [3] log
    @ ./special/log.jl:267 [inlined]
  [4] _broadcast_getindex_evalf
    @ ./broadcast.jl:709 [inlined]
  [5] _broadcast_getindex
    @ ./broadcast.jl:682 [inlined]
  [6] getindex
    @ ./broadcast.jl:636 [inlined]
  [7] macro expansion
    @ ./broadcast.jl:1004 [inlined]
  [8] macro expansion
    @ ./simdloop.jl:77 [inlined]
  [9] copyto!
    @ ./broadcast.jl:1003 [inlined]
 [10] copyto!
    @ ./broadcast.jl:956 [inlined]
 [11] copy
    @ ./broadcast.jl:928 [inlined]
 [12] materialize
    @ ./broadcast.jl:903 [inlined]
 [13] radiative_transfer(α::Matrix{Float64}, S::Matrix{Float64}, spatial_coord::Vector{Float64}, μ_points::Int64, spherical::Bool; include_inward_rays::Bool, α_ref::Vector{Float64}, τ_ref::Vector{Float64}, I_scheme::String, τ_scheme::String)
    @ Korg.RadiativeTransfer ~/.julia/packages/Korg/6TfTZ/src/RadiativeTransfer/RadiativeTransfer.jl:110
 [14] radiative_transfer
    @ ~/.julia/packages/Korg/6TfTZ/src/RadiativeTransfer/RadiativeTransfer.jl:73 [inlined]
 [15] #radiative_transfer#1
    @ ~/.julia/packages/Korg/6TfTZ/src/RadiativeTransfer/RadiativeTransfer.jl:61 [inlined]
[...]

PR #381, opened 2024-12-02
throw an error if interpolated atmosphere has negative optical depths

ajwheeler:
Partially addresses #378


ajwheeler:
Test failures here seem to be julia nightly, not me.

PR #386, opened 2024-12-17
change default

ajwheeler:
>    [Wheeler et al. 2024](https://ui.adsabs.harvard.edu/link_gateway/2024AJ....167...83W/doi:10.3847/1538-3881/ad19cc)
    demonstrates that this fits real data better, despite not being physically self-consistent.

Motivated by challenges fitting the apogee solar spectrum with the current defaults.  
image

I had originally wanted to have dissolution on, because its consistent with the way the continuum is calculated, but it's causing significant model-data mismatch (see above).

To be clear, this is a change of default behavior, which could always be controlled with a keyword argument to `synthesize`.

### DIFF
--- a/src/synthesize.jl
+++ b/src/synthesize.jl
@@ -63,8 +63,10 @@ solution = synthesize(atm, linelist, A_X, 5000, 5100)
   - `cntm_step` (default 1): the distance (in Å) between point at which the continuum opacity is
     calculated.
   - `hydrogen_lines` (default: `true`): whether or not to include H lines in the synthesis.
-  - `use_MHD_for_hydrogen_lines` (default: `true`): whether or not to use the MHD occupation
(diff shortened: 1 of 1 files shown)


ajwheeler:
@andrew-saydjari votes merge

andycasey:
May I propose:
- Immediate update to emit a warning if someone is synthesizing any wavelength between (1.5-1.7 microns) and is using the existing default behaviour to alert anyone who is fitting APOGEE/JWST spectra and not following this GitHub repo.
- That warning should note that the default behaviour will change in 1.0.
- In 1.0 change the default behaviour and remove the warning message.

andrew-saydjari:
I am in favor of this. Offline, I proposed that we either merge or do something like what Andy proposed, just so that people don't keep doing the "wrong" thing. Emitting this warning solves that problem (for me, as the person who needs to make sure I am not doing the "wrong" thing).

PR #411, opened 2025-04-02
add checks for common problems with the observational data being fit

ajwheeler:
Checks for non-finite numbers, zero'ed uncertainties, and non sorted wavelengths in`Korg.Fit.fit_spectrum` and throws more informative errors in these cases.


ajwheeler:
I need to make this work correctly with the `windows` kwarg before merging.

But the current test failures aren't that, they are actually https://github.com/JuliaMath/Interpolations.jl/issues/611.

ajwheeler:
This PR also reorganizes the tests a bit, wrapping the continuum absorption tests in a `@testset`, and moving the partition function tests to the statmech `@testset`.

PR #412, opened 2025-04-15
compat with ForwardDiff 1.0

ajwheeler:
This addresses some of the CI failures in #411 which arise from changes with the 1.0 of ForwardDiff.  Waiting on https://github.com/JuliaMath/Interpolations.jl/pull/614 to finish, test, and merge this.


ajwheeler:
These turned out not to be the biggest issue with FD 1.0, but all are good changes, and the NaN fixes are required.

PR #413, opened 2025-04-16
Threading

ajwheeler:
This PR introduces thread parallelism to speed up parts of the calculation.  In the past, my philosophy was that Korg should be entirely serial, because you can always parallelize over stars/spectra.  However, it's still nice to have a single synthesis go faster on your laptop, and multithreading makes it easier to saturate the cores on a node without bumping into the memory ceiling.

[See here for how to set the number of julia threads](https://docs.julialang.org/en/v1/manual/multi-threading/) (it's `-t` or `$JULIA_NUM_THEADS`).  If you do nothing, you will continue to get the serial behavior which should have the same performance as before.

### DIFF
--- a/src/line_absorption.jl
+++ b/src/line_absorption.jl
@@ -42,22 +42,18 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
     end
 
     # TODO
-    tasks_per_thread = 2 # customize this as needed. More tasks have more overhead, but better
+    tasks_per_thread = 1 # customize this as needed. More tasks have more overhead, but better
     # load balancing
     chunk_size = max(1, length(linelist) ÷ (tasks_per_thread * Threads.nthreads()))
     linelist_chunks = partition(linelist, chunk_size)
-
-    α_lock = ReentrantLock()
+    #TODO
+    #linelist_chunks = [linelist]
 
     tasks = map(linelist_chunks) do linelist_chunk
         # Each chunk of your data gets its own spawned task that does its own local, sequential work
         # and then returns the result
         Threads.@spawn begin
-            #lb and ub are the indices to the upper and lower wavelengths in the "window", i.e. the shortest
-            #and longest wavelengths which feel the effect of each line
-            # TODO does it make sense to preallocate these????
-            lb = 1
-            ub = 1
+            α_task = zeros(eltype(α), size(α))
 
             # preallocate some arrays for the core loop.
             # Each element of the arrays corresponds to an atmospheric layer, same at the "temps" array and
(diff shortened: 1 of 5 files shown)


src/line_absorption.jl
```diff
@@ -44,58 +43,73 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
...
-        # doppler-broadening width, σ (NOT √[2]σ)
-        σ .= doppler_width.(line.wl, temps, m, ξ)
-
-        # sum up the damping parameters.  These are FWHM (γ is usually the Lorentz HWHM) values in 
-        # angular, not cyclical frequency (ω, not ν).
-        Γ .= line.gamma_rad
-        if !ismolecule(line.species)
-            @. Γ += nₑ * scaled_stark.(line.gamma_stark, temps)
-            Γ .+= n_densities[species"H_I"] .* scaled_vdW.(Ref(line.vdW), m, temps)
+    chunk_size = max(1, length(linelist) ÷ (tasks_per_thread * Threads.nthreads()))
```
mabruzzo, line 46: Let `n_chunks` denote `tasks_per_thread * Threads.nthreads()`.

As currently written, the code only splits `linelist` into `n_chunks` iff `length(linelist)` is an integer multiple of `n_chunks`. In the more common case where `length(linelist) % n_chunks > 0`, `linelist` will be split into `n_chunks+1`. In the limit where each chunk is the same amount of work, one of of your threads will have an extra task.

The following tweak ensures that you always split `linelist` into `n_chunks`, when `length(linelist) % n_chunks > 0`. (As you're probably aware, this is modification is cheap since most compilers know how to group nearby divisions and remainder operations into a single instruction).

```suggestion
    n_chunks = tasks_per_thread * Threads.nthreads()
    chunk_size = max(1, length(linelist) ÷ n_chunks + (length(linelist) % n_chunks > 0))
```

mabruzzo (approved):
Neat!

I made a quick suggestion for marginally better partitioning.

I would also strongly recommend that you add a note (maybe just a sentence to the docstring and perhaps to the general documentation) telling people that results may not be bitwise identical run-to-run when using multiple threads (due to the nonassociativity of floating-point addition). Obviously, this only becomes a problem if lines-widths are broad compared to the range of line-centers per chunk. Even if this seems totally obvious to you/me, I really think its worth mentioning. It can totally catch people by surprise

src/line_absorption.jl
ajwheeler: Great catch, thanks!

ajwheeler:
Added a note about reproducibility to the FAQ.  An explanation of multithreading is also in #416, which I will merge after releasing an update.

PR #424, opened 2025-05-16
Add Input Isotopic Abundance Ratios

pacargile:
Added isotope information to Line object, a new input for synthesize that defines isotopic abundances, and a new function to correct log(gf)s for these isotopic abundances.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Korg"
 uuid = "acafc109-a718-429c-b0e5-afd7f8c7ae46"
 authors = ["Adam Wheeler <adamwhlr@gmail.com>", "Matthew Abruzzo <matthewabruzzo@gmail.com>"]
-version = "0.45.0"
+version = "0.46.0"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"


ajwheeler:
Thanks for the PR!  I don't have any huge problems with this approach, (In fact, I think it might be worth generalizing very slightly to allow for other kinds of adjustments). But, I'm a bit worried about any performance penalty to making the `Line` object bigger.  

I'm working on #429, which I can augment with some more targeted benchmarks.  This will hopefully help to make the performance picture clear.

src/linelist.jl
```diff
@@ -13,6 +13,7 @@ struct Line{F1,F2,F3,F4,F5,F6}
     # either γ_vdW [s^-1] per electron (as the first element, with -1 as the second) or (σ, α) from
     # ABO theory
     vdW::Tuple{F6,F6}
+    isotopes::Any
```
mabruzzo, line 16: In an ideal world, we would prefer that this is a concrete type, but I realize that's impossible. (I've got some rough ideas on working around this, but it would involve introducing a bunch of infrastructure -- which probably isn't worth doing)

In any case, we **probably** don't want this to be `Any`. I suspect that Julia implements `isotopes::Any` as a pointer to a (reference-counted) object on the heap.[^1] That means you will probably have a heap-allocation for every Line instance (it could do something clever when `isotopes` is `nothing`) and a pointer-indirection any time you want to access a stored integer value and 2 levels of indirection when `isotopes` is a vector.

If Julia implements `Union{Int, AbstractVector{<:Int}, Nothing}` and `Union{Int32, Int64, AbstractVector{<:Int}, Nothing}` as discriminated unions, one of those choices is *probably* better. In that scenario, there would be fewer heap-allocations (and fewer pointer-indirections). The cost is that `Line` is slightly larger (probably another 8 bytes). If all this is correct, I suspect that making  `Line` marginally larger will be better than the indirections.

(Hopefully, I've mad it clear that this is fairly speculative)

[^1]: it's unclear to me where information about the underlying type is stored (it could be stored within `Line` or within the heap-allocated object).

ajwheeler: If I merge this, it will definitely be after switching this to be a concrete type.

The decision is really between handling isotopes in the Line type (easier, potentially more flexible, could be generalized as "line tags"), or handling in the Formula type.  I'd be interested in any thoughts you have on that tradeoff.

ajwheeler: The benchmarks aren't posting a comment due to a permissions thing that I need to work out (I'm being conservative b/c there are potentially some "pwn request" vulnerabilities with this kind of thing).  But, you can see the results in the logs.  For syntheses without isotope info, time and allocs are surprisingly unaffected, which means that the compiler must be doing something smart in the `nothing` case.  

This definitely makes me more optimistic about merging some for of this.

PR #433, opened 2025-06-09
Reduce the number of allocs when calculating line window indices

ajwheeler:
- make sure all the members of `Korg.Wavelengths` have concrete types
- don't construct arrays in the implementation of `searchsortedfirst`/`searchsortedlast` for `Wavelengths`
- correct a method precedence problem that was causing the specialized error for constructing an empty `Wavelengths` from every being raised.

### DIFF
--- a/src/wavelengths.jl
+++ b/src/wavelengths.jl
@@ -9,6 +9,9 @@ a vector of upper and lower bounds. For example,
     Korg.Wavelengths(5000, 5500, 0.01) # explicitly specify the spacing (0.01 Å is the default)
     Korg.Wavelengths([(5000, 5500), (6000, 6500)])
 
+Wavelengths can be specified in either Å or cm. Values >= 1 are assumed to be in Å and values < 1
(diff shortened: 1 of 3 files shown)


src/wavelengths.jl
```diff
@@ -152,7 +161,10 @@ function Base.Sort.searchsortedfirst(wls::Wavelengths, λ)
     if λ >= 1 # convert Å to cm
         λ *= 1e-8
     end
```
mabruzzo, line 163: I know I'm stating the obvious, but you may want to make an issue to make sure you deal with this before the 1.0 release (or at least better document the behavior)

ajwheeler: This is intended to be a convenience, since constructing wavelengths doesn't happen often the performance is not a problem.  Thanks for catching that it wasn't documented.

ajwheeler:
Both comments have been addressed.  Thanks!

PR #438, opened 2025-07-15
Magg

### LINKED ISSUE #425: Change default solar abundances
I'm thinking Magg

ajwheeler:
closes #425, but see the discussion there.  It's not obvious to me what the default solar abundances should be.


ajwheeler:
This has a measurable (negative but small) effect on benchmarks! Presumably because more there's more oxygen and thus many molecular features are stronger.

PR #439, opened 2025-07-15
Add .git-blame-ignore-revs and ignore the commit where I did a global reformat

ajwheeler:
GitHub will use this automatically (I think?), and you can configure your local repo to use it with 
```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```
which will edit your `.git/config` file.


ajwheeler:
(It works)

PR #441, opened 2025-07-17
combine README and docs index.md into a single file

ajwheeler:
The README.me is now copied over when building the docs.  The only difficulty with this approach is that links in this document can use the Documenter.jl link resolving system (`@ref`).  This was already a constraint of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,18 @@
+# Korg.jl
+
 [![Documentation](https://img.shields.io/badge/Documentation-blue.svg)](https://ajwheeler.github.io/Korg.jl/stable/)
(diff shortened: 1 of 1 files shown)


mabruzzo:
You should put `# Korg.jl` at the top of the README.

You should also consider writing a short paragraph listing some of the features that Korg provides. For example: atmosphere interpolation, fitting narrow windows, auto-differentiation, customizing solar abundances. You don't need to go into detail, just let people know you can do these things.

It might be nice to highlight speed, but that gets hairy

ajwheeler:
Good idea.  I've listed several things. Ideally these would all be links, but I want to merge the tutorial notebooks with the docs, and that will change many of the relevant URLs.

PR #449, opened 2025-07-25
fix docstring format

mabruzzo:
Fixes formatting of a docstring


ajwheeler:
Changing the target branch because `default_solar_abundances` doesn't have a docstring on main, only `v1.0`

ajwheeler:
Sorry, didn't read this closely.  `isotopic abundances doesn't have a docstring on any branch, but we should fix that.

mabruzzo:
I was just trying to fix the formatting

PR #476, opened 2025-08-08
Fix TODO workflow

ajwheeler:
My TODO workflow has recently started failing, see e.g. #450.


ajwheeler:
This seems to have fixed it.

PR #496, opened 2025-09-05
Refactor CI/CD

ajwheeler:
An irritating failure of my format action in #437 lead me to refactor the workflows generally.

This is not a change in features, except that I am now using https://github.com/julia-actions/julia-format, which works slightly differently than my custom job.


ajwheeler:
The formatter suggestions via reviewdog are failing in the same way that they were in my custom formatter job.  It's definitely something to fix or replace, but the format check itself is working fine, so I'm going to proceed with this.

Issue #467, opened 2025-07-29
Switch to nonlinear least squares instead of Optim


ajwheeler:
https://docs.sciml.ai/NonlinearSolve/stable/solvers/nonlinear_least_squares_solvers/

Probably also upgrade `NLsolve.jl` in statmech when doing so.

PR #502, opened 2025-09-12
Add a kwarg for the method used to solve chemical equilibrium equations, and set it to `adaptive`

ajwheeler:
This seems to be strictly better than newton. Eventually, we should probably switch to a less neglected solver, probably SimpleNonlinearSolve, which can be done at the same time as #467.

@zachway I think this solves much of the atmosphere troubles.

### DIFF
--- a/src/statmech.jl
+++ b/src/statmech.jl
@@ -179,9 +179,13 @@ function solve_chemical_equilibrium(temp, nₜ, absolute_abundances, neutral_fra
     nₑ, neutral_fractions
 end
 
+# I think this ftol might be tighter than it needs to be, since both methods are converging to 
+# identical solutions in tests. Loosening is probably a good avenue to getting convergence more
+# broadly, but it's not clear to me why ftol=1e-8 produces identical solutions, so I'm hesitant to 
+# mess with it.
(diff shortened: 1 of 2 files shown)


zachway:
Howdy Adam, when you say this "solves much of the atmosphere troubles", are you referring to the non-convergence of low-temperature models in the plot below?

image

--Zach

zachway:
Hi Adam, I reran the BOSS grid and this branch seems to have fixed things in some place, but maybe stopped working in others. The first figure here is the original grid from the `main` branch. The second figure is the same grid run with the `solver-method` branch. Things get much better down to 3200 kelvin but the lowest temperature start failing at a higher rate. Overall convergence is higher in this temperature regime.
grid_frac_4250K
grid_frac_branched

ajwheeler:
The two methods are yielding bitwise identical results, at least for these params:
```julia
nₜ = 1e15
nₑ_initial_guess = 1e12
T = 5700
```
I'm slightly confused by this.  I don't expect it based on the `ftol` I'm using, but I've confirmed that the traces are different for the two methods, so it's clearly that I'm having a thinko about how the equations are specified and how the `ftol` translates to per-species-number density tolerance.  This is likely an area where convergence could be improved.

ajwheeler:
@zachway did this work for you?

Issue #431, opened 2025-06-05
depth-dependent convective velocity shifts


ajwheeler:
requested by @pacargile.  Maybe also useful for @palumbom.

I think the way to approach this is to have a velocity shift per layer of the atmosphere.

pacargile:
@palumbom As part of developing our new grid of ATLAS-12 model atm's, we are interested in introducing a simplistic, depth-dependent treatment of convective velocity shifts. Bob Kurucz measured some empirical shifts as a function of tau_ross for Arcturus and the Sun, and these do seem to improve the line shapes at R > ~100K with SYNTHE. I have some ideas about how to parameterize these shifts beyond these two stars (sort of similar to the Reiners et al. work), but these shifts would just be included as an additional columns in our atmosphere models. (I have also been thinking about parameterizing the dispersion of convective velocities which could be used as a depth-dependent broadening term).

I dug into SYNTHE to see how Bob had implemented his velocity shifts, and it looks very similar to what you suggested in option 3, i.e., he applied a per layer velocity shift during the RT. At this point, I had only really been thinking about doing something similar, but I totally agree there are better/more realistic ways of shifting the line shapes (e.g., your option 2). Having said that, perhaps starting simple would be an effective approach -- particularly if you already have code that does this sort of per layer velocity shifts (and broadening)?

PR #533, opened 2026-04-07
per-atmospheric layer doppler shift

varghesereji:
Added a feature to pass a velocity profile to synthesise the function.

### DIFF
--- a/src/synthesize.jl
+++ b/src/synthesize.jl
@@ -1,4 +1,5 @@
 using Interpolations: linear_interpolation
+using Interpolations
 import .ContinuumAbsorption: total_continuum_absorption
 using .RadiativeTransfer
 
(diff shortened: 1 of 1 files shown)


mabruzzo:
Just an FYI, the choice to linearly interpolate absorption coefficients (rather than use cubic interpolation) was intentional.

While cubic interpolation is higher order, it also has drawbacks. Namely, it doesn't necessarily preserve monotonicity of the data and can artificially overshoot your interpolation grid.

But we should wait to hear from @ajwheeler (maybe his thoughts have changed)

ajwheeler:
I think that commit message was referring to the interpolation that @varghesereji introduced in his first commit, so not change our continuum opacity interpolator.  However, I'm very curious as to whether you found the cubic spline to be necessary, or if it was more of a "this is probably better" thing.

Thank you for opening this PR!  I think the general approach is definitely worth implementing. To merge it, I'd need to ask for a few changes. There are some minor merge conflicts, and I'd want to leave some line-by-line comments. I'm happy to do that, but only if you have the time and interest. You will not offend me if you'd rather not do any more.

related: #431

ajwheeler:
I'm gonna close this for now for my own organization, but feel free to reach out/reopen if you like. Thanks!

PR #544, opened 2026-05-07
Mol pressure broadening

zachway:
This pull request fundamentally changes how `Line` is structured so that the pressure broadening for molecules can be calculated. Most of the major changes are summarized in `GAMMA_MOL_LORENTZ_IMPLEMENTATION.md` (thanks copilot)


zachway:
I realized that an exponent was negative... smh

zachway:
Sorry, the exponent is correct.

.CondaPkg/.gitattributes
ajwheeler, line 1: You'll want to add `.CondaPkg` to `.gitignore` and then remove these files from the PR. (Kinda surprised it's not in there already.)

data/ExoMol_header_data/states_header_data.txt
ajwheeler, line 1: Ideally we would have a README in this folder with a one sentence (or more) explanation of what the data is.

data/EXOPLINES_broadening_coeffs/table_3.txt
```diff
@@ -0,0 +1,19 @@
+Absorber Broadener a0 a1 a2 a3 b1 b2 b3 b4
```
ajwheeler, line 1: Likewise for this one.

src/linelist.jl
```diff
@@ -1,30 +1,34 @@
-using CSV, HDF5, LazyArtifacts
+using CSV, HDF5, LazyArtifacts, DataFrames
 using Pkg.Artifacts: @artifact_str
 
 #This type represents an individual line.
-struct Line{F1,F2,F3,F4,F5,F6}
+#The Line type has two broadening modes:
+# Mode 1 (default): gamma_stark and vdW broadening (backward compatible)
+# Mode 2 (optional): gamma_mol_lorentz with temperature exponent n_exp
+struct Line{F1,F2,F3,F4,F5,F6,F7}
```
ajwheeler, line 8: While I don't see a technical problem with this (i.e. I think it should work fine), it unfortunately builds on a kinda bad system I had going on.  I think it's fine to use it like this to generate data for now, but someone (can be me, can be you) should refactor before merging.

Calling all the current implementations of pressure broadening by neutral perturbers "van der Waals" was already a bit of an abuse of notation. I would like to drop it and introduce an abstract `NeutralPerturberPressureBroadeningParameters` (open to shorter names) class with implementations for (1) the "normal" gamma_vdW per perturber at 10_000 K (2) ABO params and (3) this implementation.  (Idk if it make sure to further split up your implementation, I read enough of the diff yet.)

This will hopefully make the code a lot easier to follow.  We'll have to benchmark to make sure we don't introduce more run-time indirection that we already had, but I think that should be doable.

ajwheeler: (in light of think, I'm going to skip over giving truly nitty-gritty feedback on this `Line` code.)

src/linelist.jl
```diff
@@ -182,6 +253,39 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
...
+
+Calculates the Padé equation approximation of the broadening coefficient (in s^-1 bar^-1).
+This prescription is to be used strictly for estimating the broadening per Gharib-Nezhad (2021).
+"""
+function gamma_pade(J::Union{AbstractFloat, AbstractVector{<:AbstractFloat}}, 
+                    A::AbstractVector{<:AbstractFloat}, 
+                    B::AbstractVector{<:AbstractFloat})
+    # Unpack coefficients
+    a_0, a_1, a_2, a_3 = A # ← A must be a vector/array!
+    b_1, b_2, b_3, b_4 = B # ← A must be a vector/array!
```
ajwheeler: Comments here are prob not helpful.

src/linelist.jl
```diff
@@ -182,6 +253,39 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
...
+
+    # Calculate γ_L per Gharib-Nezhad (2021) equation 4
+    γ_L = (a_0 .+ a_1.*J .+ a_2.*J.^2 .+ a_3.*J.^3)/(1 .+ b_1.*J .+ b_2.*J.^2 .+ b_3.*J.^3 .+ b_4.*J.^4) # in cm^-1 bar^-1
+
+    # convert cm^-1 atm^-1 to cm^-1 bar^-1
+    # could be considered cm^-1 if we multiply by p/p_ref in line_absorption.jl where p_ref = 1 bar
```
ajwheeler: This one (second line only) I'd argue is confusing.

src/linelist.jl
```diff
@@ -182,6 +253,39 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
...
+
+function return_γ_b(J, absorber, broadener)
+    exoplines = CSV.read(joinpath(_data_dir, "EXOPLINES_broadening_coeffs", "table_3.txt"), DataFrame, delim=" ")
+    exoplines_species = Korg.Species.(exoplines.Absorber)
+    row = (absorber .== exoplines_species) .&& (broadener .== exoplines.Broadener)
+    γ_b = gamma_pade(J, 
+                     vec([exoplines.a0[row][1], exoplines.a1[row][1], exoplines.a2[row][1], exoplines.a3[row][1]]),
+                     vec([exoplines.b1[row][1], exoplines.b2[row][1], exoplines.b3[row][1], exoplines.b4[row][1]]))
+    # in cm^-1 bar^-1
+    # could be considered cm^-1 if we multiply by p/p_ref in line_absorption.jl where p_ref = 1 bar
```
ajwheeler: ditto

src/linelist.jl
```diff
@@ -182,6 +253,39 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
     γstark, log_γvdW
 end
 
+"""
+    gamma_pade(J, A, B)
+
+Calculates the Padé equation approximation of the broadening coefficient (in s^-1 bar^-1).
+This prescription is to be used strictly for estimating the broadening per Gharib-Nezhad (2021).
```
ajwheeler: can you add a doi.org or ADS link for this paper?

src/linelist.jl
```diff
@@ -206,6 +310,8 @@ Load a linelist from ExoMol. Returns a vector of [`Line`](@ref)s, the same as [`
   - `T_line_strength`: the temperature (K) at which to evaluate the line strength (default: 3500.0)
   - `isotopic_abundances`: the table of isotopic abundances to use (default: `Korg.isotopic_abundances`).  This is ignored if `isotopic_correction` is provided.
   - `verbose`: if `true` (default), will print progress information
+  - `broad_files`: the location of the ExoMol .broad files for H2 and He. This is only used if this variable is set and only works with H2O currently.
```
ajwheeler: It's not super clear what "this" refers to in "This is only used".

Is it true that this only works for H2O or is that out of date?

src/linelist.jl
```diff
@@ -206,6 +310,8 @@ Load a linelist from ExoMol. Returns a vector of [`Line`](@ref)s, the same as [`
...
+  - `path_flag`: integer value that determines what "path" of pressure broadening to follow. 0) no pressure broadening, 1) use ExoMol's .broad files, and 2) use the EXOPLINES approximations
```
ajwheeler, line 331: Can you change this so that the parameter is a string specifying the mode instead?

src/linelist.jl
```diff
@@ -232,18 +340,87 @@ function load_ExoMol_linelist(spec, states_file, transitions_file, lower_wavelen
         @info "The states and transitions files, $states_file and $transitions_file, don't contain the string 'states' or 'trans', respectively. You may have mixed them up."
     end
 
+    if path_flag == 2
+        if spec ∉ Korg.Species.(["AlH", "CaH", "MgH", "CrH", "FeH", "TiH", "SiO", "TiO", "VO"])
```
ajwheeler, line 360: Ideally these would not be hard coded.

src/linelist.jl
```diff
@@ -232,18 +340,87 @@ function load_ExoMol_linelist(spec, states_file, transitions_file, lower_wavelen
...
+            throw(ArgumentError("EXOPLINES approximation for pressure broadening not yet implemented for this species"))
+        end
+        if broad_files != nothing
+            throw(ArgumentError("Trying to set broadening parameters for species with both EXOPLINES and ExoMol prescriptions. For the following species, ExoMol broadening from the .broad files is not supported, but is for the EXOPLINES prescription: AlH CaH MgH CrH FeH TiH SiO TiO VO"))
+        else
+            J_ = vec([J for J in 0:0.5:500])
```
ajwheeler, line 368: Can you add comment saying this is the total angular momentum quantum number (assuming it is)?

src/linelist.jl
```diff
@@ -232,18 +340,87 @@ function load_ExoMol_linelist(spec, states_file, transitions_file, lower_wavelen
...
+                broad_He = CSV.read(broad_file, DataFrame, delim=" ", silencewarnings=true, ignorerepeated=true, header=false)
+            else
+                throw(ArgumentError("He or H2 are not the final characters before .broad"))
+            end            
+        end
+
+        broad_H2[ismissing.(broad_H2.Column5), "Column5"] .= -1
+        broad_He[ismissing.(broad_He.Column5), "Column5"] .= -1
+
+        # Use both J and K quantum numbers. See Table 1 of https://www.homepages.ucl.ac.uk/~ucapsy0/pdf/18BaHiCz.pdf
```
ajwheeler, line 397: Can you add this link too: https://doi.org/10.1016/j.jqsrt.2017.01.028 ?
Feel free to keep the non-elsevier one as well though.

src/linelist.jl
```diff
@@ -296,7 +478,35 @@ function load_ExoMol_linelist(spec, states_file, transitions_file, lower_wavelen
         f = A * prefactor * g_upper / (g_lower * wavenumber^2)
         log_gf = log10(g_lower * f) + isotopic_correction
         E_lower = Korg.hplanck_eV * wn_lower * Korg.c_cgs
-        line = Korg.Line(1.0 / wavenumber, log_gf, spec, E_lower)
+        if path_flag == 2
+            # n_exp set to 0.5 for all species given Gharib-Nezhad et al. 2021
+            # also leave in cm^-1
+            line = Korg.Line(1.0 / wavenumber, log_gf, spec, E_lower; 
+                                gamma_mol_lorentz=(γ_H2[J_lower], #^-1 * (Korg.c_cgs * 4π) * wavenumber^2, 
+                                                   γ_He[J_lower]), #^-1 * (Korg.c_cgs * 4π) * wavenumber^2),
```
ajwheeler: Comments should be deleted.

src/line_absorption.jl
```diff
@@ -87,9 +88,28 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
...
+                    kboltz_pressure = 1.380649e-22 # in cm^3 * bar / K
+                    # Sum over perturbers: H2 and He
+                    # gamma_total_lorentz still in cm^-1 here
+                    @. gamma_total_lorentz = line.gamma_mol_lorentz[1] * (T_ref / temps)^(line.n_exp[1]) * (n_densities[species"H2"] * kboltz_pressure * temps / p_ref)  # H2 contribution
+                    @. gamma_total_lorentz += line.gamma_mol_lorentz[2] * (T_ref / temps)^(line.n_exp[2]) * (n_densities[species"He"] * kboltz_pressure * temps / p_ref)  # He contribution
+                    # sum in lorentz and convert to frequency s^-1
+                    @. Γ += gamma_total_lorentz * c_cgs * 4π #gamma_total_lorentz^-1 * (c_cgs * 4π) / line.wl^2
+                else
+                    # Mode 1: Normal broadening as per the Korg paper
+                    if !ismolecule(line.species)
```
ajwheeler: Should be an elseif?

src/line_absorption.jl
```diff
@@ -87,9 +88,28 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
                 # sum up the damping parameters.  These are FWHM (γ is usually the Lorentz HWHM) values in
                 # angular, not cyclical frequency (ω, not ν).
                 Γ .= line.gamma_rad
-                if !ismolecule(line.species)
-                    @. Γ += nₑ * scaled_stark.(line.gamma_stark, temps)
-                    Γ .+= n_eff_vdW .* scaled_vdW.(Ref(line.vdW), m, temps)
+                if !ismissing(line.gamma_mol_lorentz) && !ismissing(line.n_exp)
+                    # Mode 2: Molecular Lorentz broadening with temperature dependence
```
ajwheeler, line 92: I don't think we should refer to "mode 1, mode 2" out of context here.  But this will probably be irrelevant after my proposed changes to `Line`.

src/line_absorption.jl
```diff
@@ -87,9 +88,28 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
...
+                    # Sum contributions from H2 (index 1) and He (index 2)
+                    # Scale as: gamma * (T_ref / T)^n_exp
+                    # in Barton et al. 2017 eqn 1 as well as Gharib-Nezhad et al. 2021 eqn 2
+                    # (with opposite signs for the exponent)
```
ajwheeler: Why are signs opposite?   Ideally should explain, even if the answer is just "convention".

src/line_absorption.jl
```diff
@@ -87,9 +88,28 @@ function line_absorption!(α, linelist, λs::Wavelengths, temps, nₑ, n_densiti
...
+                    T_ref = 296 # Kelvin
+                    p_ref = 1 # bar
+                    kboltz_pressure = 1.380649e-22 # in cm^3 * bar / K
```
ajwheeler: Would it make sense to convert to Barye (cgs pressure) when constructing the `Line` so we could use `kboltz_cgs` (already defined) here?

src/molecular_cross_sections.jl
```diff
@@ -31,6 +31,8 @@ this function, though they can be saved and loaded using [`save_molecular_cross_
     at which to precompute the cross-section.
   - `log_temp_vals` (default: 3:0.025:5): The log10 of the temperatures at which to precompute the
     cross-section.
+  - `log_perturber_density` (default: 12:0.5:20): The log10 of the pertruber density at which to precompute the
```
ajwheeler: typo

src/molecular_cross_sections.jl
```diff
@@ -31,6 +31,8 @@ this function, though they can be saved and loaded using [`save_molecular_cross_
...
+    cross-section. This is split into 85% H2 and 15% He.
```
ajwheeler:
```suggestion
    cross-section. A mix of 85% H2 and 15% He is assumed.
```

src/molecular_cross_sections.jl
```diff
@@ -91,24 +96,27 @@ function Base.show(io::IO, sigma::MolecularCrossSection)
...
-                                               number_densities) where R<:Real
+                                               number_densities,
+                                               perturber_densities) where R<:Real
     if length(molecular_cross_sections) == 0
         return
     end
 
     for sigma in molecular_cross_sections
-        # this is an inefficient order in which to write to α, but doing the interpolations this
-        # way uses less memory.
```
ajwheeler, line 108: Can you add this comment back?

src/synthesize.jl
```diff
@@ -209,7 +209,7 @@ function synthesize(atm::ModelAtmosphere, linelist, A_X::AbstractVector{<:Real},
     abs_abundances ./= sum(abs_abundances) #normalize so that sum(n(X)/n_tot) = 1
 
     #float-like type general to handle dual numbers
-    α_type = promote_type(eltype(atm.layers).parameters..., eltype(linelist).parameters...,
+    α_type = promote_type(eltype(atm.layers).parameters..., Float64,
```
ajwheeler, line 212: This will break some autodiff calculations.  I think the original version should work?

src/synthesize.jl
```diff
@@ -270,7 +270,8 @@ function synthesize(atm::ModelAtmosphere, linelist, A_X::AbstractVector{<:Real},
                                               Korg.Wavelengths([atm.reference_wavelength * 1e8]),
                                               get_temps(atm),
                                               vmic,
-                                              number_densities)
+                                              number_densities,
+                                              number_densities[species"H2"] .+ number_densities[species"He I"])
```
ajwheeler: Seems a bit weird to pass both of these.

GAMMA_MOL_LORENTZ_IMPLEMENTATION.md
```diff
@@ -0,0 +1,126 @@
+# Implementation Summary: Tuple-Based Molecular Lorentz Broadening for `Line` Type
```
ajwheeler, line 1: Also should be removed before merging

Project.toml
```diff
@@ -9,6 +9,7 @@ projects = ["test", "benchmark", "docs"]
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
ajwheeler, line 12: Ideally we would drop this dependency, but we will need to add bounds to it if we keep it.

test_exomol_H2O.ipynb
ajwheeler, line 1: notebooks also should be removed before merging.

src/linelist.jl
```diff
@@ -182,6 +253,39 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
...
+
+    # Calculate γ_L per Gharib-Nezhad (2021) equation 4
+    γ_L = (a_0 .+ a_1.*J .+ a_2.*J.^2 .+ a_3.*J.^3)/(1 .+ b_1.*J .+ b_2.*J.^2 .+ b_3.*J.^3 .+ b_4.*J.^4) # in cm^-1 bar^-1
+
+    # convert cm^-1 atm^-1 to cm^-1 bar^-1
+    # could be considered cm^-1 if we multiply by p/p_ref in line_absorption.jl where p_ref = 1 bar
+    γ_L / 1.01325
+end
+
+function return_γ_b(J, absorber, broadener)
```
ajwheeler:
```suggestion
function load_γ_b(J, absorber, broadener)
```

src/linelist.jl
```diff
@@ -182,6 +253,39 @@ function approximate_gammas(wl, species, E_lower; ionization_energies=ionization
...
+    exoplines = CSV.read(joinpath(_data_dir, "EXOPLINES_broadening_coeffs", "table_3.txt"), DataFrame, delim=" ")
```
ajwheeler: Ideally we would not use DataFrames. I would like to drop that as a dependency.

ajwheeler (changes requested):
There aren't any correctness problems that are clear to me, so you think it's working, then I see no problems with using it for calculations for now. (With the caveat that I haven't read the relevant papers and am trusting you on that one.)

The comments and things that I would like to address before merging, which can take longer. They span a large level of pickiness.  

High-level feedback:
- I use a formatter to keep all the formatting consistent.  (That's the reason for one of the failing checks)
- there are a couple broken tests, shouldn't be too bad to fix.
- Ideally, we would add a few to ensure that this functionality doesn't get broken in the future.

src/linelist.jl
zachway: I agree this should be refactored. 

The implementation could be split up in a few different ways. 1) there are different prescriptions for when the broadening is actually measured (as it is with H2O) versus when it is modeled (as with the EXOPLINES species). 2) The broadening relations are different for H2 and He and these could be plug-and-play if we wanted to do that. I think number 2 is a much lower priority than 1.

src/line_absorption.jl
zachway: I made this change and converted the lorentz coefficient to cm^-1  barye^-1

src/linelist.jl
zachway: The EXOPLINES prescription for broadening only works for these nine species, so they should probably be listed somewhere. I changed the error to better describe this.

src/linelist.jl
zachway: Done

src/linelist.jl
zachway: I adjusted the language here. 

It is only implemented for H2O right now because that is the only `.states` file header data that I tabulated in `data/ExoMol_header_data/stated_header_data.txt` for which the pressure broadening is measured. Other species that could be added are NH3, SO2, CH4, PH3, HCN and H2CO.

src/line_absorption.jl
zachway: That makes sense but I'll leave this open for now

src/line_absorption.jl
zachway: Added some text here. The problem is that the equations from these papers are defined pretty differently even though they are mathematically the same thing.

src/molecular_cross_sections.jl
zachway: Yes, I don't know why copilot removed that. Spooky...

src/synthesize.jl
zachway: I agree this is weird but this allows `interpolate_molecular_cross_sections!` to take both number densities and perturber_densities as arguments separately. These are important to keep separate because of the way they interact with `MolecularCrossSection`. 

This could be refactored in some way but I'd need to think harder about it

ajwheeler:
Lmk when you are ready for me to take another look at this :)

PR #550, opened 2026-05-24
New chemical equilibrium

ajwheeler:
There was a bug in the core chemical equilibrium solver that lead to resultant pressure/number density not matching those from the model atmospheres when there were a significant number of molecules. This fixes that, ~~which required adding another parameter to the system of equations.~~

Internal changes:
- removed `solve_chemical_equilibrium`, which was an unnecessary layer.
- Korg now solves the system of equations in terms of log number densities, rather than number densities
- H- is now self-consistently incorporated into the system of chemical equilibrium equations.  This doesn't make it meaningfully more accurate, but it's conceptually simpler, and it's nice to be able to get H- number densities out of Korg more easilly.
- If the initial solve fails, the new solver modifies the system of equations to "turn down" the effect of molecules until it gets to convergence, then "anneals" back to the correct system ("parameter continuation").
- I've replaced the call to NLsolve with a hand-rolled newton solver. This is mostly about ergonomics.


ajwheeler:
closes #516

PR #574, opened 2026-08-26
Simplify continuum absorption bounds checking

ajwheeler:
(This is prep for incorporating the changes in #563)

Eliminates the `bounds_checked_absorption` machinery that was used for a subset of continuum sources. It was limited utility since it wasn't used for everything and it forced a function signature that wasn't always right. The robust bounds checking code is intact. It's just used directly by a subset of the continuum opacity sources.

The modified functions are now more idiomatic and have become type-stable (I think).


mabruzzo:
FWIW, this looks fantastic to me! It seems like a clear win!

At the time that I wrote this, it wasn't yet clear to me that the continuum opacities are relatively unimportant (beyond getting you into the right ballpark). I suspected that if continuum opacity was important, there would be interest in customizing the continuum opacity, and that people might would therefore want to invoke our continuum opacity functions outside of Korg for testing purposes. Thus, providing a polished semi-standard interface for bounds checking seemed useful. I think this was also before I understood, that the limitations of certain kinds of opacities (e.g. metal-opacities) would have very different kinds of bounds. I also never thought that was really an optimal way to create a standard interface for bounds-checking (I had thoughts about how to do it, but wasn't sure how best to do it Julia). In other words, this was all quite aspirational and I'm quite happy to see this cull.

ajwheeler:
Thanks for putting your eyes on it!

A note for anyone else reading this: the continuum opacities are quite important, but people tend not to be interested in modifying Korg's defaults, which is a use case we initially had in mind as more central.

PR #579, opened 2026-09-02
add H2-H2 and H2-He CIA

ajwheeler:
This is adapted from Phill's PR (#563), which ports several continuum opacity sources from SYNTHE. I'm merging them one at a time to facilitate review.

This PR adds collision induced absorption from H2-H2 and H2-He interactions. See https://adsabs.harvard.edu/pdf/1997A%26A...324..185B for an intro. It is a particularly important opacity source for low metallicity M dwarfs (cc @zachway).

This plot gives you an idea of the size of the effect. Note that the top panel is not a realistic spectrum because molecule lines have been left out.
image


mabruzzo:
> This plot gives you an idea of the size of the effect. Note that the top panel is not a realistic spectrum because molecule lines have been left out.

That's insane!

